PR #39, opened 2018-10-19
Dem/Shape functions

Zeitsperre:
The goals of this PR are to create classes/methods for the following:

- User create an instance of a spatial object with a dem and a basin contour as a shapefile/GML/WKT document.

- The class takes the dem and shape/contour and returns the following:
    - Mean elevation of the basin; 
    - Basin centroid and total area; 
    - Eventually, land-use cover, slope, and others.

Area and slope (and centroid) will be the most tricky as these demand an appropriate CRS transformation but can be implemented with with Proj4 (`PyProj`). Relevant CRSes for Quebec span UTM zones 17 to 21 or MTM zones 7 to 10.

I'm currently writing this out the hard way (with `osgeo` tools like `gdal` and `ogr`) since I haven't landed on a good library for working with both raster and vector data (present options include: `fiona, raterio, PyProj, Shapely, (Geo)Pandas`). Once all the boilerplate code is written and I can manage to get some functionalities working, it should be easy to replace methods with better alternatives once I've landed on them.


huard:
I suggest splitting the vector part and the raster part. Shapely works well for vector stuff (https://gis.stackexchange.com/questions/127607/area-in-km-from-polygon-of-coordinates)

Zeitsperre:
Thanks for the suggestion. I think I will probably write this class as more shape-focused and add some general methods for passing it a DEM that returns the averaged values (slope, elevation, etc.) of the DEM.

I'm reluctant to build a class that relies on a ton of libraries, but `osgeo`, `pyproj` are absolutely necessary. It's currently setup for multi-feature geometries but I imagine the shapes being passed to a working implementation of this class would be single features (as WKT objects)?

Zeitsperre:
Clipping and averages from clipped segments seem to be working. Clipping is based on lat/lon extent of features. Still need to add some identifying information (FIDs of features). 

Will look further into clipping according to cutline; It might involve casting the shape to an array with the same grid as DEM and using it as a mask with numpy.

huard:
@Zeitsperre  Is it ready to review ?

Zeitsperre:
@huard Worked on this today and wanted to give you an update. 

What's going well:
- Only uses low level OSGEO libraries and fairly fast as a result.
- Current working calls for `RavenShape` are `centroids()`, `clip(raster)` (to features), and `average(raster)` (based on output from `clip`).
- Helper methods for returning the corners and bbox of shape/raster. Projection checking is performed on file import and can be done against EPSG codes directly.

What needs to be done:
- The clipping methods are based on boundary boxes; Rasterio can provide easy masking based on feature cut-lines. Looking into this tomorrow.
- Slope calculations can be performed using GDAL subprocess. Will look into creating methods that perform the calculation and masking. Can also use the `check_crs` to ensure that CRS is projected before performing slope. 
- Area calculations require reprojection methods in the RavenGIS object (Easy to do in GDAL).
- The RavenRaster object will be used for Land Use data that shouldn't be reprojected. Should have methods for pixel/area statistics (categories and frequencies) with option of loading a shape to get feature statistics (checks CRSes --> reproject shape --> clip --> stats).

I can work on this for a few more days this week as well. Please feel free to review and comment if anything needs clarifying.

Zeitsperre:
Found out today that rasterio is an "all-or-nothing" deal. Converting from GDAL to rasterio will have to wait. https://rasterio.readthedocs.io/en/latest/topics/switch.html

Zeitsperre:
@huard Feel free to review now. Slope is still not implemented but I can aim for that tomorrow. Definitely needs a bunch of refactoring but would really appreciate your thoughts on its current state. Thanks!

raven/gis.py
```diff
@@ -0,0 +1,472 @@
...
+    @staticmethod
+    def raster_bbox_corners(raster):
+        """Returns the corner coordinates from a raster grid"""
+        upper_left_x, x_res, x_skew, upper_left_y, y_skew, y_res = raster.GetGeoTransform()
+
+        lower_right_x = upper_left_x + (raster.RasterXSize * x_res)
+        lower_right_y = upper_left_y + (raster.RasterYSize * y_res)
+        return upper_left_x, upper_left_y, lower_right_x, lower_right_y
+
+    @staticmethod
```
huard: xd, yd are the width and height, correct ? I think this should be documented. It's not just a coordinate transformation.

raven/gis.py
```diff
@@ -0,0 +1,472 @@
...
+
+    def __str__(self):
+        return 'RavenGIS Raster: {}'.format(self._filename)
+
+    def projected(self):
+        if self._raster_reproj is not None:
+            return self._raster_reproj
+        return False
+
+    def stats(self):
```
huard: Stats should probably return a dict rather than print to stdout.

huard: I mean, you could rename this function to print_stats, which would call stats.

Zeitsperre: I actually like the first option better. I'm not a fan of printing directly within a class either. I'll get to this soon.

huard:
@Zeitsperre  Is that PR still relevant ?

Zeitsperre:
@huard Nope. The classes and methods written here are a bit stale and I've saved the more interesting (to me) code snippets that I might integrate into personal projects down the line.

PR #45, opened 2018-11-09
Generic raven

huard:
## Overview

Big refactor to simplify the wrapping of models and also launching models independently of the WPS process.


huard:
@julemai I've refactored the code to simplify adding new emulators. Ideally you would review it and provide comments.

julemai:
@huard @richardarsenault 

Sorry, I wasn't in office beginning of the week. I had a quick look too and I'm not sure if the 1D NetCDF inputs (only dimension is time) would work. I actually doubt it.

David, would it be possible to give me a quick introduction on how I can test the WPS codes on my machine? I need to run that to reliably see if it is working or not.

huard:
Normally, you should be able to do 
```
$ make install 
$ make test
```
but since the URL to the zip of the code is from a previous version, you have to copy the latest raven  executable (tested with 153) manually in bin/raven after the `make install`.

Let me know if you hit a snag.

huard:
I just created a new branch with the src included. This way we'll be able to just make install and it should work.

julemai:
Hi David @huard,

I resolved that issue with providing the source code at another [download location](http://www.civil.uwaterloo.ca/jmai/raven/raven-rev157.zip). Shall I continue working on the branch `generic_raven` or `include_src`?

julemai:
Hi David @huard,

revised the code. Let me know if that looks good to you. I had to change quite a bit.

Have a great weekend,
Julie

huard:
@julemai Looks good. I'll implement something to take care of derived parameters on the server side rather than having the user have to do it.

julemai:
@huard Sounds good! For sure. The derived parameters should be created automatically. Doesn’t the implemention i did already did that? 

I should ha e mentioned derived parameters earlier. Also, that parameters are not only in the rvp file but can be in all of the 5 raven setup files.

huard:
@julemai From what I understand, users calling the emulators still had to assign derived parameter values themselves (as you did in the unit tests). I added a method in each Model subclass to compute derived parameters from the main parameters (see `derived_parameters` in basic.py).

julemai:
Hi David @huard,

I see. The additional function is just perfect. thanks!!

huard:
Great !
@Zeitsperre The tests are failing because `make install` is not running on travis ( if my diagnostic is correct). The makefile includes the build of the raven executable (`raven_dev`). Is there a reason to not just call `make install` in the travis config ?

Zeitsperre:
@huard No, calling `make install` should be fine. Unless we want to leverage `tox`, we can simply add that to the `before_install` or `install` sections.

PR #49, opened 2018-11-20
Travis CI

Zeitsperre:
Adding some of the Travis CI configurations here. Won't be employing `tox` since we just want it to build a few ways.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ language: python
 python:
   - "2.7"
   - "3.6"
+branches:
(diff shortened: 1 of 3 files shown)


Zeitsperre:
Travis now builds for PRs against master. Some `AssertionError` and `TypeError` failures differ between Python versions so some compatibility checking needs to be done. Lots of PEP8 to deal with as well. Are we interested in having 3.7, 3.5, 3.4 builds?

PR #50, opened 2018-11-20
PEP8

Zeitsperre:
Gradually cleaning up a lot of the code. Will refactor the variables and classes to follow best practices. Present changes deal with quite a bit.

### DIFF
--- a/raven/models/hbv.py
+++ b/raven/models/hbv.py
@@ -1,10 +1,11 @@
 import numpy as np
 import scipy.signal as ss
 
-def simulation(data, params=[ 1.0,   0.15,     250,   0.055,\
-                            0.055,   0.04,     0.7,     3.0,\
-                              1.5,    120,     1.0,     0.0,\
-                              5.0,    0.7,     0.05,    0.1]):
+
+def simulation(data, params=[1.0, 0.15, 250, 0.055,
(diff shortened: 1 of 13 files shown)


raven/models/hbv.py
```diff
@@ -1,10 +1,11 @@
...
 
-def simulation(data, params=[ 1.0,   0.15,     250,   0.055,\
-                            0.055,   0.04,     0.7,     3.0,\
-                              1.5,    120,     1.0,     0.0,\
-                              5.0,    0.7,     0.05,    0.1]):
+
+def simulation(data, params=[1.0, 0.15, 250, 0.055,
+                             0.055, 0.04, 0.7, 3.0,
+                             1.5, 120, 1.0, 0.0,
+                             5.0, 0.7, 0.05, 0.1]):
```
Zeitsperre: This is a dangerous practice as on multiple calls, the parameters can be changed without user knowledge.

huard:
So the non Raven models is code from another repo that i put there to test
the WPS interface. I'm not sure we should invest time to clean it up. I'm
sorry i did not make that clear earlier. So i suggest you commit your
cleanup, but that we archive these models for the time being.

Le mar. 20 nov. 2018 16:35, Trevor James Smith <notifications@github.com> a
écrit :

PR #51, opened 2018-11-26
Implement make install within Travis CI

Zeitsperre:
Trying to implement `make install` in Travis CI environment.


Zeitsperre:
@huard I've managed to make it work using the `make` processes. Lessons learned:
- WPS-based processes appear to be broken. There is probably a way to specify open ports that can be called within travis, but this would need to be specified either within the `.travis.yml` (as originally specified) or with the `makefile` somewhere.
- Using `make test` as the evaluation script will not fail the build even if there are failed tests. This is because if the environment can run `make test` successfully, this returns 0. Not very ideal.
- For some reason, the Travis base system (native Python3.6 in Ubuntu Xenial) installs miniconda3 (natively Python3.7) which then proceeds to install Python3.6 even though it seem to not be specified anywhere. This seems a bit redundant. Most libraries should work in Python3.7. Investigating this further.

julemai:
Hm. Works for me. Just removed the `src/RAVEN.zip` the Makefile is looking for and it downloads the code, unzips, and compiles:

```
Downloading RAVEN hydrological framework ...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  526k  100  526k    0     0  1584k      0 --:--:-- --:--:-- --:--:-- 1585k
Unzipping RAVEN ...
Compiling RAVEN ...
Installing application ...
```

What's your error message?

Zeitsperre:
The error message is in the travis-ci check:
```
Compiling RAVEN ...
make[1]: Entering directory '/home/travis/build/Ouranosinc/raven/src/RAVEN'
make[1]: *** No targets specified and no makefile found.  Stop.
make[1]: Leaving directory '/home/travis/build/Ouranosinc/raven/src/RAVEN'
Makefile:79: recipe for target 'raven_dev' failed
make: *** [raven_dev] Error 2
The command "make install" failed and exited with 2 during .
```

Zeitsperre:
I get the impression that the new `raven_dev` model is missing the makefile or maybe the makefile is elsewhere. It seems to be crashing on this line:
```
	@test -f $(RAVEN_SRC)/raven_rev.exe || $(MAKE) -C $(RAVEN_SRC) -j4
```

julemai:
Can you please go to that folder and run the make manually:

```
cd /home/travis/build/Ouranosinc/raven/src/RAVEN
make
```

julemai:
Could it be that you don't have NetCDF installed on your system. I mean the netCDF library in general, not the python NetCDF...

Zeitsperre:
Figured it out. The error is because the new folder created a folder structure and the makefile was deeper. Calling unzip with `-j` fixes this.

julemai:
Sorry about that. Wasn't causing any issues on my side. I not even realized that it was different.

Zeitsperre:
All good. That seems to have fixed that issue. The same problems I mentioned above are again valid. I might consider forgoing `make test` and revert to activating conda in the script. This would also remove the need to `before_install` flake8. Getting the wps processes to play nicely can come next.

Zeitsperre:
This looks about done on my end. The build is failing due to the pyproj library being inaccessible at the moment. In addition, it seems to be building on both Linux and macOS.

Zeitsperre:
closes #49

PR #52, opened 2018-11-27
Added examples of tests looking into the data output files.

huard:
## Overview

Example of how the model output can be opened to compare results against actual known values. 

Let me know if something is missing.


julemai:
Hi David @huard,

I'm done I think. It was super easy with the functions you've provided. :)

PR #56, opened 2018-11-29
Travis tweak

huard:
## Overview

- Name for builds. 
- Separate build for PEP8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ branches:
 sudo: false
 before_install:
   - if [ $TRAVIS_OS_NAME = linux ]; then sudo apt-get install libnetcdf-dev; else brew install netcdf; fi
-  - if [ $PEP8 = true ]; then sudo apt-get install flake8; fi
+  - if [ $PEP8 = true ]; then sudo apt-get install python-flake8; fi
 install:
   - make install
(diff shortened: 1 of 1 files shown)


Zeitsperre:
I can take it from here. ;)

huard:
Ok, that's my last try then I'll leave it to you.

huard:
I installed travis locally, and there is a `travis lint` bash script to check the validity of travis config files.

Zeitsperre:
Just noting here but the travis ruby tool doesn't install very cleanly in some machines due to the deprecation of `rubygems` (ie: broken in cutting-edge Linux distros). Travis CI is releasing a new version soon.

Zeitsperre:
Pretty happy with the results here. @huard Looking at the C compiler problems makes me think we could consider collecting the RAVEN source code and making a build for macOS and Linux from source. Perhaps not now but down the line.

julemai:
Sorry I lost a bit track here. Which kind of C compiler problems do you have? Is it due to Raven? I just know that distributing an executable is usually hard when NetCDF is included.

huard:
Another option is to use a docker image of the build. Marc Lalonde has already prepared something, I need to take a look.

PR #58, opened 2018-12-03
adding everything for MOHYSE

julemai:
@huard I have incorporated MOHYSE. Mostly working but not last test (see below). Could you have a look?

`tests/test_basic.py::TestRaven::test_mohyse PASSED`
- yields same result as template in Wiki

`tests/test_wps_raven.py::TestRavenProcess::test_mohyse PASSED`

`test_wps_raven_mohyse.py::TestRavenMOHYSEProcess::test_simple is FAILING`

- probably due to parameters that need to get changed in RVH file besides the standard name, area, elevation etc.
- error message:
```
  File "/Users/j6mai/Documents/GitHub/raven/raven/models/basic.py", line 386, in derived_parameters
    self.rvh['par_rezi_x10'] = 1.0 / self.rvh['par_x10']
TypeError: unsupported operand type(s) for /: 'float' and 'NoneType'
```


julemai:
Well, you couldn't have know before. I should have told you. Technically parameters can be in any (of the five Raven) files. I guess it would be cleaner to have the separated in rvh, rvp, and rvc. Otherwise I think the code won't be that readable.

Let me double-check if the last test is matching the template in the Wiki before you merge.

I can do HBV-EC then tomorrow or the day after.

huard:
I guess my question boils down to : 
Do we need to support parameters that have the same name but different values in different rv files ? So for example x1 in rvp and x1 in rvc that have different values. 
If not, we can simplify the backend by just having one big RV dictionary where we store all inputs. Since all names would be different, we won't have any name clashes. I'll need to make a few changes to support our magic_parameters, but that should not be too bad.

julemai:
I see. No! They will need to have always different names. Otherwise we will face problems in the OSTRICH calibration as well.

If there will be parameters that have the same name AND are in different files (never seen such an example yet), we can always create a derived one by x1_new = x1 * 1.0.

What do you think?

huard:
Sounds good. I suggest you go ahead and merge that branch once you're satisfied the tests run, and then I'll create another PR for the RV refactoring which hopefully will simplify all this.

julemai:
@huard How can I find the model setup that was used for the test RAVEN runs? It seems that the folders get deleted. At least I can't find them anymore. I need those files to find out why model runs are not matching the expected output. Previously I was just printing the path:

`self.model_path` and `self.output_path` after line 214 in `basic.py` and then I was able to track them down. But now they don't exist anymore...

huard:
I think those temp folders are deleted at the end of the test. So when the
test fails, they remain on disk, when the test completes, they're deleted.
If you're using a debugger, you can add a breakpoint during execution and
then inspect workdir to get the path and inspect the files.
Another option down the line would be for us to return the rv files as
outputs, so you could inspect them as well from the process response.

julemai:
I tried to force a crash of the test. That and right before printing the tmp-folders but that isn't working anymore. Saving the rv*-files will be helpful because we might want to safe them when we do calibrations. For now. The runs are thankfully all matching the wiki templates. I will do one final commit that has all tests PASSING and then try to merge... Never done that before on git... Let's see how that goes.

PR #60, opened 2018-12-05
Raven HBV-EC

julemai:
Is mostly done. The only thing that isn't working is the last test `TestRavenHBVECProcess.test_simple` because parameters `par_x20` and `par_x21` need to get filled into *.rvt file. 

See `basic.py` line 414.

@huard Could you please have a look? When this is fixed I will make sure that the tests are yielding the same results as the templates in the Wiki. THANKS!!


julemai:
Two more things we need to do (mostly noting them here for myself):

1. **HBV-EC specific** <br>HBV requires the specification of `MonthlyAveEvaporation` and `MonthlyAveTemperature` in the `*.rvt` file. Those values need to be derived from the inputs. Right now they are fixed.
2. **General for all settings** <br>The *.rvt file also contains the `Latitude`, `Longitude`, and `Elevation`. These values need to get filled in automatically as well. Right now, they get filled in only in the `*.rvh` file.

huard:
@julemai I've been working on a refactoring of the RV class to simplify all this. One question I have is about the handling of parameters and derived parameters. I guess OSTRICH will require us to define a parameter array that includes the derived parameters, correct ? So we need a mechanism with input parameters (user defined), and output parameters (user + derived) that would feed into both the RV templates and the OSTRICH config ?

julemai:
@huard In OSTRICH there will be two blocks as well. One for the original parameters (the one you called user defined) and the a second block for derived parameters (called tied parameters in OSTRICH). Here is an example copied from an OSTRICH setup with 21 original model parameters and two that get derived:

```
#Parameter/DV Specification
BeginParams
  #parameter      init.    low      high    tx_in  tx_ost  tx_out   # in HBV called
  par_x01         random  -3.0       3.0    none   none     none    TT 
  par_x02         random   0.0       8.0    none   none     none    CFMAX
  par_x03         random   0.0       8.0    none   none     none    CFR
  par_x04         random   0.0       0.1    none   none     none    CWH
  par_x05         random   0.0       1.0    none   none     none    par_x5 = FC/par_x21
  par_x06         random   0.3       1.0    none   none     none    LP
  par_x07         random   0.0       7.0    none   none     none    BETA
  par_x08         random   0.0     100.0    none   none     none    PERC
  par_x09         random   0.01      1.0    none   none     none    K1
  par_x10         random   0.05      0.1    none   none     none    K2
  par_x11         random   0.01      6.0    none   none     none    MAXBAS
  par_x12         random   0.0       5.0    none   none     none    PCALT
  par_x13         random   0.0       5.0    none   none     none    TCALT
  par_x14         random   0.0       0.2    none   none     none    saturation at the wilting point 
  par_x15         random   0.0       1.0    none   none     none    ALPHA
  par_x16         random   0.0      30.0    none   none     none    maximum interflow rate for capillary rise
  par_x17         random   0.01      3.0    none   none     none    thickness of top soil layer
  par_x18         random   0.0       2.0    none   none     none    melt correction factor (forest)
  par_x19         random   0.05      1.0    none   none     none    release from glacier as it melts
  par_x20         random   0.8       1.5    none   none     none    RFCF
  par_x21         random   0.8       1.5    none   none     none    SFCF
EndParams

BeginTiedParams
  # par_1_+_x15 = 1.0 + par_x15
  # Xtied = (c1 * X1) + c0
  # --> c0 = 1.0
  # --> c1 = 1.0
  #
  par_1_+_x15 1 par_x15 linear 1.00 1.00 free
  #
  # par_half_x11 = par_x11 * 0.5 
  # Xtied = (c1 * X) + c0
  # --> c0 = 0.0
  # --> c1 = 0.5
  #
  par_half_x11 1 par_x11 linear 0.5 0.0 free
EndTiedParams
```

This is the setup for HBV and I derived those two parameters in the PAVICS Raven as well.

Does that help?

huard:
Yes ! Just to be 100% clear, that means that we don't need to worry about derived parameters for OSTRICH, it will handle those internally ?

julemai:
That's correct! It will be handled in the second block by OSTRICH itself.

julemai:
We only need to provide template files (similar to the PAVICS RAVEN) and OSTRICH will look for the pattern matching, e.g., `par_1_+_x15` and will fill in the value it has derived before.

huard:
ok good. What I'll do is finish the RV refactoring, merge it into master, then merge master into your hbv branch and find a solution for the mean computations feeding the x20 and x21 parameters.

julemai:
@huard Sounds good! 

Parameters `x20` and `x21` are not the `MonthlyAveEvaporation` and `MonthlyAveTemperature`.  They need to get additionally filled in the *.rvt file:

```
[...]
:Latitude               54.4848
:Longitude             -123.3659
:Elevation              843.0     

:RainCorrection         {par_x20}   # HBV_PAR_20 == RFCF
:SnowCorrection         {par_x21}   # HBV_PAR_21 == SFCF

#                       Jan    Feb    Mar    Apr    May    Jun    Jul    Aug    Sep    Oct    Nov    Dec
:MonthlyAveEvaporation, 0.028, 0.089, 0.315, 1.003, 2.077, 2.959, 3.190, 2.548, 1.382, 0.520, 0.101, 0.023,
:MonthlyAveTemperature, -11.4, -7.2,  -2.8,  2.8,   8.1,   12.2,  14.4,  13.5,  8.9,   3.4,   -4.0,  -9.3,
[...]
```

huard:
Thanks for pointing that up, I got confused.

huard:
@julemai Could you take a look ? I've just added the mean monthly temp and evap to the computation of derived parameters. The values I compute seem to match those you have, but since you rounded and I'm not, the NSE is not identical so the test fails. 

I ended up creating a rvd for derived parameters, otherwise it broke the existing logic in the code.

huard:
What I do is insert a breakpoint in the code (I use pycharm for development). Then once inside the program, I can inspect temporary objects and the temp folder created by the process (self.workdir). 

I understand what you'd like is to inspect the temporary folder that is created during process execution. I know we can do that when the server is really running (make start), but I don't know how to do that with pytest. I'll ask around. Another option is to output the configuration files, but this will take me some time to do correctly as I'm not familiar with the standard to do this (metalink). So apart from the breakpoint solution, I have no quick fix on offer...

julemai:
Hm. That's strange because it was working before. I don't want to use any other development tool to do the debugging. I don't want to get used to these tools and then they are not available on clusters anymore. But there must be something in the code that cleans up the run-folders, isn't it? Otherwise how can the model run in a folder that doesn't exist anymore at the end?

huard:
OK I understand. Have your tried to run with `pytest --pdb`. It should stop at errors and then enter into the trace. You can then probably do `print(self.workdir)` and then cd into that directory. 

Just asked on https://gitter.im/bird-house/birdhouse
We use the pytest library. It's probably written somewhere in the docs. Let me know if you find it.

julemai:
I just tried `pytest`. It seems that it does not find any library:

```
[j6mai@JulieUW:/Users/j6mai/Documents/GitHub/raven/]:
pytest --pdb
=========================================================================== test session starts ===========================================================================
platform darwin -- Python 3.7.0, pytest-4.0.1, py-1.7.0, pluggy-0.8.0
rootdir: /Users/j6mai/Documents/GitHub/raven, inifile: setup.cfg
collecting ... 
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> traceback >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
ImportError while importing test module '/Users/j6mai/Documents/GitHub/raven/tests/test_LHMP_models.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/test_LHMP_models.py:1: in <module>
    import numpy as np
E   ModuleNotFoundError: No module named 'numpy'
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> entering PDB >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /Users/j6mai/miniconda/lib/python3.7/site-packages/_pytest/python.py(482)_importtestmodule()
-> "{traceback}".format(fspath=self.fspath, traceback=formatted_tb)
(Pdb) 
```

Any idea?

huard:
So my colleagues tip is to call 
```
import time
time.sleep(600)
```
which will wait for 600 seconds. If you print the temp directory, that should give you some time to look into the folder.

Zeitsperre:
@huard neat workaround!

julemai:
@huard Loading the python environment using `source /Users/j6mai/miniconda/bin/activate raven` helped. It now finds everything. But it still cleans up the folders.

The point is that I get all the outputs just after all tests finished (no matter if I use `make test` or `pytest`. So, at that point everything is already cleaned up. The sleeping time in between doesn't help.

julemai:
I just killed the `pytest` manually when it was sleeping and that didn't delete the folder. Not nice but worked.

julemai:
Done. Test is now passing. 

I've updated also the example templates on the Wiki accordingly.

PR #61, opened 2018-12-08
Refactor RV handling

huard:
## Overview

Refactors the RV class to provide more flexibility in how parameters are specified and assigned to rv files.

Changes:

* Parameter arrays are now stored as namedtuple. 
* RV objects can include both a parameter array and normal parameters.


huard:
No ; ) It was late...

PR #66, opened 2018-12-17
spotpy installation fix

huard:
## Overview
Continuation of #65

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -10,5 +10,6 @@ dependencies:
 - psutil
 - xarray
 - pandas
-- spotpy
 - bumpversion
+- pip:
+  - spotpy
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,3 +9,4 @@ netCDF4
 dask
 toolz
 pandas
+spotpy


Zeitsperre:
Yeah, this is dealt with in the rasterio branch as well as in the cookiecutter. The issue is hyphens in the bash call for the makefile for testing. I'll fix it here right now.

julemai:
The weights of all time steps is always 1.0 in Raven. Could it be that the time series are shifted by one day??

julemai:
Ohhh. And you need to make sure that data are not NODATA. I can look into that tomorrow if you like. 

Please let me know where exactly the SpotPy is calculating the RMSE.

huard:
I'm skipping the first time index as does RAVEN, then spotpy just does `return np.sqrt(mse(evaluation, simulation))` where `mse` is:
```
        obs, sim = np.array(evaluation), np.array(simulation)
        mse = np.mean((obs - sim)**2)
        return mse
```

julemai:
Guessing you need to use `np.nanmean()`.

julemai:
The `nanmean()` is also not working. I will look at that later tonight or tomorrow. Have to leave now.

huard:
Excellent ! Thanks. I'm having trouble on my end however. The fourth digit does not seem to match. 

Run `flake8 raven tests`

@richardarsenault there are a few issues I'd like you to think about. The process uses hard-coded variable names (q_sim, q_obs). That's fine as long as we're only using Raven generated files, but we could add optional LiteralInputs to allow users to specify the variable they want. 

Also, the output is a JSON dictionary keyed by metric name and storing the information for the first basin. If we have a netCDF file with multiple watersheds, how do you want to store the output ? Should we create a netCDF file for the output ? Note that we could have both types of outputs (json and netCDF).

huard:
That's it. I'll create a "raven_clean" make command to do the clean-up.

julemai:
Sounds good. The makefile has a `make clean`. But I'm sure you already saw that. Thanks for the `pep8` command. I'll run that next time before I commit.

richardarsenault:
@huard 
I think this code will be used exclusively (or almost exclusively) by either (1) Us, in development mode and to feed our other algos such as for calibration, displays on figures, etc.; and (2) users that want to do calibrations/simulations and get that info. In both cases, we control the input/output and postprocess them for our needs. What I mean is that I don't think anyone is going to call Pavics directly to compute objective functions with their local Qobs/Qsim data. So in that light, I think we can keep it simple and leave it in the current JSON format.

Also, I still get an assertion error between RAVEN and SPOTPY RMSE? Maybe you can see if the Raven_clean method will clear this up?

Finally I think that we should keep it simple and only do one basin at a time (no multi-catchment). If a user (or one of our processes) wants to do multi-basin, we should simply iterate and call this function N times.

julemai:
@richardarsenault Where do you get an error? When I run the tests in this branch, they all go through without any failure.

Happy New Year!

huard:
My guess is that you have an older version of the RAVEN code. I just pushed the raven_clean make command.

richardarsenault:
Hi, yes, it's probably the case. I still have the version I installed like a month ago. Still getting used to this GitHub/branching thing, especially in command line. Sorry, hope the objective-function code works well though.

PR #68, opened 2019-01-01
Added regionalisation branch manually...

richardarsenault:
Local branch push to origin is not pushing files, uploading them manually. Sorry for the large PR.

## Overview

This PR adds the regionalization tools, including: 
- WPS Regionalization process
- Test WPS regionalization
- 7 regionalization methods
- Outputs a netcdf file containing the nDonor simulations as well as a best estimate of the true streamflow
- Databases for pre-calibrated watersheds for three hydrological models (GR4JCN, HMETS and MOHYSE)
- Random table of properties for the gauged catchment database, which will need to be modified when the geo tools will be available

ALSO:
- Some modifications to wpsio and other support packages to add the regionalization process
- Addition of some packages required during deployment (spotpy)

TODO / ISSUES: 
1- MOHYSE is not running properly in this branch. Something to do with a mandatory parameter par_X08 that it seems is not available, although I am passing the parameters.
2- GR4J is having difficulty running with 2+ donors (parameter sets), something to do with RAVEN crashing due to "bad data". I think maybe parameter sets are being refused by RAVEN, maybe do to combinations of parameters that lie outside the feasible range??
3- Clean the import sections, I added too many when I was trying to debug and just copy/pasted whole blocks to make sure I didn't forget anything.
4- Transform the regionalization_tools.py and associated datasets to csv format instead of the current .txt format
5- Update the geo data in the gauged_basins_properties.txt file
6- Was unable to import "common" from the wps, so made a copy in the processes folder. Probably not the best solution but I can't find how to do it otherwise.
7- some paths are absolute and need to be changed to relative (i.e in regionalization_tools.py). Again, I tried and can't make it work.
8- Some files can be deleted (RegionalizationHydrograph.nc, entrypoint_regionalization.py)


Makefile
```diff
@@ -3,7 +3,7 @@ APP_ROOT := $(CURDIR)
 APP_NAME := raven
 
 # Anaconda
-ANACONDA_HOME ?= $(HOME)/miniconda
+ANACONDA_HOME ?= $(HOME)/anaconda3
```
huard: I'm concerned this will break other people's code.

raven/processes/__init__.py
```diff
@@ -5,6 +5,11 @@
 from .wps_raven_hmets import RavenHMETSProcess
 from .wps_raven_hbv_ec import RavenHBVECProcess
 from .wps_objective_functions import ObjectiveFunctionProcess
+try:
+    from .wps_regionalisation import RegionalisationProcess    
+except ImportError:
```
huard: the try except should be removed.

raven/processes/entrypoint_regionalization.py
```diff
@@ -0,0 +1,172 @@
...
+    # Create a dataframe that is identical to gauged_properties arrays.
+    gauged_properties, model_parameters = rank_donor_catchments(
+                                            ungauged_properties,
+                                            model_parameters,
+                                            gauged_properties,
+                                            properties_to_use,
+                                            regionalization_tag
+                                            )
+
+    if regionalization_method == 1:
```
huard: I suggest giving names to the different methods.

raven/processes/regionalization_methods.py
```diff
@@ -0,0 +1,326 @@
...
+           Qsim[0:Qs.shape[0],j]=Qs.flatten()
+        
+       else:
+           print("The hydrological model selected is not supported for \
+                 regionalization")
+
+    return Qsim
+
+
+def runLocalHMETS(params,ncfilename,start_date, end_date, latitude, longitude, elevation, area ):
```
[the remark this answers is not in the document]
richardarsenault: just saw your docs, indeed it'll be easier this way... I spent a lot of time making this WPS-WPS connection work.

raven/processes/entrypoint_regionalization.py
richardarsenault: OK will do that soon.

Makefile
richardarsenault: Yes I had to do this to run on my machine, I'll remove it.

raven/processes/__init__.py
richardarsenault: OK will do

richardarsenault:
@huard 
when you say binary files, etc., you mean the compiled raven code and such? I am guessing spopty too? I am guessing I can leave the Database files (the parameters, properties, etc.)?

raven/processes/regionalization_methods.py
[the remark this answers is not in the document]
huard: Sorry about that. I'll see if I can update the run method to flatten the argument list (pass start_date directly instead of inside the rvi dictionary.).

Makefile
huard: I think you can set your own environment variable to override this default.

Zeitsperre:
If the .gitignore is being followed in the repo, performing a `git clean` should automatically remove logs and `.pyc` files. For an example of how that looks, check out https://stackoverflow.com/a/36573710/7322852

richardarsenault:
Closing, Replaced by PR #76

PR #76, opened 2019-01-11
Improved regionalisation PR

richardarsenault:
## Overview

Fixed PR #68 errors and problems. 

Still has MLR to fix when databases will be updated.

Changes:


raven/processes/regionalization_methods.py
```diff
@@ -0,0 +1,260 @@
...
+            
+        elif hydrological_model == "MOHYSE":
+
+            Qs=runLocalMOHYSE(params,ncfilename,start_date, end_date, latitude, area )
+            Qsim[0:Qs.shape[0],j]=Qs
+            
+        else:
+            print("The hydrological model selected is not supported for regionalization")
+ 
+    Qsim=np.delete(Qsim,list(range(Qs.shape[0],Qsim.shape[0])),axis=0);
```
huard: Est-ce que cette ligne est juste pour prendre les n premier time step ?

raven/processes/regionalization_methods.py
```diff
@@ -0,0 +1,260 @@
...
+
+    # Preallocate empty array with final dimensions
+    nc_file = Dataset(inputs_file[0], 'r')
+    nDays=nc_file.variables["rain"][:].shape[0]
+    Qsim = np.zeros(shape=(nDays,1))
+    
+    # Make parameter vector
+    MLR_param=MLR_parameters        
+ 
+    if hydrological_model == "GR4JCN":
```
huard: Je viens de t'écrire une fonction run_model qui va simplifier la logique.

raven/processes/wps_regionalisation.py
```diff
@@ -0,0 +1,273 @@
...
+            model_name=request.inputs['model_name'][0].data
+        else:
+            raise ValueError('Model Name is not optional')
+
+        if model_name not in ['GR4JCN','HMETS','MOHYSE']: 
+            raise ValueError("Hydrological model " + model_name + " is not supported for regionalization.")
+
+        #Regionalisation Method
+         
+        if 'regionalisationMethod' in request.inputs:
```
huard: Normalement avec la denière version de PyWPS, la gestion des valeurs par défaut devrait être faite pour toi automatiquement.

raven/processes/wpsio.py
```diff
@@ -122,6 +122,30 @@
                          data_type='float',
                          min_occurs=1)
 
+model_name = LiteralInput('model_name', 'Hydrological model identifier',
+                         abstract="Hydrological model indetifier: {HMETS, GR4JCN, MOHYSE}",
+                         data_type='string',
+                         default='HMETS',
+                         min_occurs=0)
```
huard: Tu peux utiliser allowed_values pour spécifier les valeurs acceptées. J'ai fait la modif dans ma branche.

richardarsenault: OK, je te laisse faire les modifs proposées de ton côté ou tu veux que je les fasse en parallèle?

raven/processes/regionalization_tools.py
```diff
@@ -0,0 +1,256 @@
...
+        # Get the values from the arrays, convert to numpy array
+        similar_properties = gauged_properties[properties_to_use].values
+        similar_targets = np.tile(
+                            ungauged_properties[properties_to_use].values,
+                            (similar_properties.shape[0], 1))
+
+        # Calculate deltas in the database
+        delta_max = np.amax(similar_properties, axis=0, keepdims=True)
+        delta_min = np.amin(similar_properties, axis=0, keepdims=True)
+        deltas = delta_max - delta_min
```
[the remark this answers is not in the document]
richardarsenault: Yes, that is an acceptable option, thanks

raven/processes/wpsio.py
huard: Je m'en occupe. Je vais te soumettre une version révisée.

raven/processes/regionalization_methods.py
```diff
@@ -0,0 +1,260 @@
...
+                                    model_parameters,
+                                    gauged_properties,
+                                    properties_to_use)
+
+    model_parameters = model_parameters.values[:, 2:]
+
+    for i in range(0, r_squared.shape[0]):
+        # if we have an R2 > 0.5 then we consider this to be a better estimator
+        if r_squared[i] > 0.5:
+            model_parameters[:, i] = MLR_parameters[i]
```
huard: I don't understand this. Why would you replace one parameter by the MLR parameter ? If the r2 is > .5 for all parameters, then you're running the model number_donor times with the same parameter set, no ? If it's not, then you have a mix of parameters from the ungauged catchments and the MLR parameters which won't necessarily work well together.

richardarsenault: Yes, the idea is that when you transfer a parameter set, you kind of implicitly "decide" that the best available info for each parameter comes from the neighboring/most similar donor. However, if you have some parameters/processes that are more easily identified with a MLR, then this info should be included. It is a philosophical point, some people will say that parameter sets should be left completely intact because the parameters are a set that works together, others will say that each one represents a physical process and should be individually determined from observations, etc. This method (regression-augmentation) is kid of the compromise solution. In this paper [http://richardarsenault.ca/Continuous_streamflow_prediction_ungauged_basins_effect_equifinality.pdf ] we showed that it performed better than all the others in many situations. So for example we could change only 3 parameters out of 21 if those are the only ones that have a certain skill with MLR.

richardarsenault:
OK corrected the gauged_properties file IDs.

raven/processes/regionalization_methods.py
huard: Interesting. I'm trying to restructure the code and avoid duplication. I should be finished sometimes today.

huard:
@Zeitsperre I'm currently refactoring. These PEP8 changes will create problems with my merge.

Zeitsperre:
Okay. Want to git revert to 92dcc8e ?

huard:
@richardarsenault OK. I pretty much rewrote the whole thing... 

- I created a regionalize function in the utilities/regionalization.py file, with all related tools inside. 
- The wps has become much simpler as it parses the inputs and calls regionalize, then assigns the outputs. 
- I've use xarray and pandas functions to simplify some of the code's logic. 
- I've moved some of the WPS inputs that are specific to regionalization inside wps_regionalization. We'll keep wpsio for reusable inputs and outputs (ie. used in more than one process). 

I think it's going to make the code easier to maintain over time, but I'm guessing it has introduced some bugs as well, which brings me to "testing". 

How are we planning to test that all the regionalization methods work as expected ? Can we create a simple synthetic property and parameter file only for testing ? 

So please take a look, check that I didn't do too many stupid mistakes (pretty sure there'll be a couple of them).

richardarsenault:
OK excellent, I'll take a look at all of this, including preparing tests to verify the code, within the next 48-72 hours. Thanks a lot for your work!

raven/utilities/regionalization.py
```diff
@@ -0,0 +1,377 @@
...
+        spread = stats.loc['std']
+    elif kind == 'iqr':
+        spread = stats.loc['75%'] - stats.loc['25%']
+
+    d = ungauged.values - gauged.values
+    n = np.abs(d) / spread.values
+    return pd.Series(data=n.sum(axis=1), index=gauged.index)
+
+
+def regionalization_params(method, gauged_params, gauged_properties, ungauged_properties, filtered_params, filtered_prop):
```
huard: Pourquoi est-ce qu'on a besoin de gauged_params et filtered_params (gauged_properties et filtered_prop) ?

richardarsenault:
All tests work, results as expected. Merging.

PR #84, opened 2019-03-11
Run multiple-model in parallel

### LINKED ISSUE #13: Add support for parallel multi-model analysis (8)
This would probably be a process calling other processes to benefit from parallelism and scalability. Not clear to me how that would work.

huard:
## Overview

This PR fixes #13 

Changes:
* Multiple outputs are zipped for the WPS output.


huard:
@julemai This PR fails on Mac and it's hard for me to debug just from the Travis output. I've changed the way raven is called to support multiple parallel processes and I guess that's where the error comes from. 

You'll see in base.py that `run` returns a list of processes, and `__call__` waits for them. My guess is that on Mac the process either doesn't start, or python doesn't wait for it to complete. 

Do you have some time to look at this ?

julemai:
> @julemai This PR fails on Mac and it's hard for me to debug just from the Travis output. I've changed the way raven is called to support multiple parallel processes and I guess that's where the error comes from.
> 
> You'll see in base.py that `run` returns a list of processes, and `__call__` waits for them. My guess is that on Mac the process either doesn't start, or python doesn't wait for it to complete.
> 
> Do you have some time to look at this ?

On it now.

julemai:
@huard Seems like the problem is that the output folder that is used in `cmd` in base.py line 371 (option -o) is not existing. It expects `<path>/model/p00/output` but `output` is not created. Wondering why this would work on other platforms though.

huard:
Thanks !

julemai:
All passing now on my Mac. :)

huard:
Yé ! @richardarsenault, you'll need to approve the PR for me to merge it.

PR #85, opened 2019-03-12
Graphic showing ensemble of simulations

### LINKED ISSUE #14: Add support for multi-model uncertainty analysis (8)
Is this a process or a workflow ?

huard:
## Overview

This PR fixes #14 

Changes:

* Added process creating a figure showing the mean annual cycle for each simulation in a zip file. 

## Related Issue / Discussion
Very, very basic figure. @richardarsenault you'll probably want to modify it. See it as a starting point.


richardarsenault:
@huard I'm looking at the code and I'll propose a few improvements / additions, such as adding some graphs that are continuous timeseries, envelopes, etc. However I can't seem to get the name of the model from the generator object in the graphing function. I'd need that to add a legend. If you can show me how to get that, then I can add a few images.

I'll also fix 1- Adding matplotlib package in the default build and 2- The WPS_caps function that is currently failing.

richardarsenault:
I've implemented a rudimentary solution for now, but @huard it would be good to check that the nc-file generator passes the info in the same order as the folder list as that is how I scrape the run names.

huard:
The model name does not seem to be included. Will add it to the generated netCDF files.

huard:
@julemai It would be nice if the RVI file contained a :Model_ID field, which would store the name of the model configuration, and write is as a global attribute in the netCDF files.

julemai:
@huard I see. Can we use ‘:RunName’ for that? I would change RAVEN such that it dumps this as global attribute to the NetCDF. Can work on that on Sunday.

huard:
@julemai I suggest we not do that, because it confuses two distinct things: the experiment name (run_name) and the model used for the experiment (model_id). I can always add those attributes via the python interface, but I think the clean solution is for Raven to store that information from the get go.

huard:
I suggest that for the moment I make a temporary fix using the python interface, and open a new ticket for a change to raven that would add a section in the rvi file called `:NCAttrs`. This section would let users set global attributes to the netCDF file. For example:
```
:NCAttrs
 :model_id raven-gr4j
 :source École de technologie supérieure
 :contact Richard
:EndNCAttrs
```
And we don't need that by Monday !

julemai:
Sounds good! Can we please call it `:NCAttributes`? And I am not sure if C++ will be able to handle fancy French accents.

julemai:
Talked to James. We will share workload on implementations in RAVEN. So, it's on the way...

julemai:
@huard Newest RAVEN version (rev 176) is out with this feature. You can have as many of the following lines in the `*.rvi` file and it will put all these as global attributes to the NetCDF file. Let me know if this is what you wanted:

```
:NetCDFAttribute [name of attribute (string)] [attribute value (string)]
```

I made the new code available again through my webpage `http://www.civil.uwaterloo.ca/jmai/raven/raven-rev176.zip`. You only need to adjust this in the makefile.

huard:
@julemai I don't seem to have permission to access that file.

julemai:
Oops, sorry @huard. Please try again. I had a typo in the file name...

huard:
Bingo ! Thanks

huard:
Has there been other changes to the code ? I'm now getting 
```ERROR : CommonFunctions: StringToCalendar: Unknown calendar specified!```

**Comment by julemai** [comment]
Yes, there's a code change. We are supporting different calendars now. Hence you need to specify the calendar attribute for the time variable. For example,
```
double time(time=20819);
  :standard_name = "time";
  :units = "days since 1954-01-01 00:00:00";
  :calendar = "gregorian";
```
We support `STANDARD`, `GREGORIAN`, `PROLEPTIC_GREGORIAN`, `NOLEAP`, `ALL_LEAP`, `365_DAY`, `366_DAY`, `JULIAN`. Is the attribute set for your NetCDF file?

**Comment by huard** [comment]
Yes, it's the same salmon river example, which uses gregorian as its calendar.

**Comment by julemai** [comment]
Arghhhh! C++ and strings. Always again a pleasure... Looking into this now. Sorry!

**Comment by julemai** [comment]
Can you please try the new revision?! 

`http://www.civil.uwaterloo.ca/jmai/raven/raven-rev177.zip`

**Comment by huard** [comment]
@julemai Is there a way for the attribute value to include spaces ? Let's say I want to store 
```
:NetCDFAttribute references  Craig, J.R., and the Raven Development Team, Raven user's and developer's manual (Version 2.8), URL: http://raven.uwaterloo.ca/ (2018).
```
How would I do that ? At the moment it only saves "Craig".

**Comment by julemai** [comment]
James did that implementation. So, I need to dig into that. But I told him that we need to allow for blanks, commas etc.

**Comment by huard** [comment]
Ok, no rush. The current implementation allows us to get back the model_id, which is what we we're looking for in the first place.

**Comment by julemai** [comment]
@huard new version dumps whole string given after the attribute name but might convert some of the delimiters (blanks, tabs, commas) to blanks

`http://www.civil.uwaterloo.ca/jmai/raven/raven-rev178.zip`

**Comment by richardarsenault** [comment]
@huard I think this one is ready to go, for now at least. Probably room for improvement but it's a good start.

File: raven/processes/wps_graph_ensemble_uncertainty.py [hunk 1]
```diff
@@ -0,0 +1,69 @@
...
+import zipfile
+from pathlib import Path
+from raven.utilities.graphs import mean_annual_hydrograph, hydrograph
+from matplotlib import pyplot as plt
+
+class GraphEnsUncertaintyProcess(Process):
+    def __init__(self):
+        inputs = [ComplexInput('sims', 'Stream flow simulations ensemble',
+                               abstract='Stream flow simulation time series',
+                               supported_formats=(Format(mime_type='application/zip'),)),
```
huard: Add the netCDF Format as well, so people can submit a list of files.

huard: also, you can use 
```
from pywps import FORMATS
supported_formats=(FORMATS.NETCDF)
```

raven/processes/wps_graph_ensemble_uncertainty.py
```diff
@@ -0,0 +1,69 @@
...
+        fig.savefig(fig_fn_simple)
+        plt.close(fig)
+    
+        
+  #      zipPath=Path(self.workdir) / 'ensemble_hydrographs.zip'
+  #      with zipfile.ZipFile(zipPath,'w') as zip: 
+            
+  #           for file in [fig_fn_annual,fig_fn_simple]:
+  #              zip.write(file) 
+  #      #TODO: ZIP FIGURE FILES TOGETHER
```
huard: Or not. Is it a problem to send back two figures ?

raven/processes/wps_graph_single_hydrograph.py
```diff
@@ -0,0 +1,87 @@
...
+from matplotlib import pyplot as plt
+
+class GraphSingleHydrographProcess(Process):
+    def __init__(self):
+        inputs = [ComplexInput('sim', 'Stream flow simulations NetCDF File',
+                               abstract='Stream flow simulation time series',
+                               supported_formats=[FORMATS.NETCDF]),
+                  ]
+
+        outputs = [ComplexOutput('graph_ensemble_hydrographs', 'Figure showing the simple hydrographs of the included models.',
```
huard: Here we have only one model right ?

huard:
Bon pour moi, tu peux merger quand tu veux.

raven/processes/wps_graph_single_hydrograph.py
richardarsenault: Yes here there is only one model, single NetCDF file containing a Qsim and Qobs.

richardarsenault: Changed the "ensemble" to "single" for this case

raven/processes/wps_graph_ensemble_uncertainty.py
richardarsenault: True, I already send those back but I thought it might be simpler to handle if we ever go to a case with many figures. I'll leave it as unzipped for now.

raven/processes/wps_graph_ensemble_uncertainty.py
richardarsenault: Done

PR #91, opened 2019-03-13
Fix 89

### LINKED ISSUE #89: Support multiple params from WPS call
## Description

At the moment parallel parameters can only be called from the Model class, not the WPS interface.

huard:
## Overview

This PR fixes #89.

Changes:

* Improved support for parallel processes driven by vectors of parameters.


richardarsenault:
@huard I'll check this PR in the next few days. If I understand correctly, this PR is for the parallel parameter set evaluations and not the parallel models or basins. And we are running the parameter sets in parallel using system calls, not PyWPS-based parallelism / job management. Correct?

huard:
Correct. One call to WPS -> multiple system calls to Raven. 
This PR is because I realized that although we could call the python interface to raven with multiple parameters, the WPS did not yet support multiple parameters.

richardarsenault:
Everything works perfectly, thanks. Merging

PR #92, opened 2019-03-15
Hardcoded PET formula for HMETS and GR4J

richardarsenault:
PET is now calculated from the OUDIN formula instead of using user-generated PET for HMETS and GR4J


richardarsenault:
@huard This should work, @julemai and I have fixed the remaining bugs.

PR #94, opened 2019-03-19
WIP - Hydrobasins optimize

### LINKED ISSUE #86: Optimize the watershed selection process
## Description

The process to find the watershed and its upstream contributors is too slow. We need to optimize it. 

One option is to let GeoServer do part of the work if possible.

Zeitsperre:
## Overview

This PR fixes #86 

Changes:

* Added lookup tables for HydroBASINS levels 7-12 with/out lakes
* Integrated tables via Pandas for the upstream basin identification process

## Related Issue / Discussion
* The selection algorithm still needs to be improved. WIP

## Additional Information


Zeitsperre:
I'm looking into the best way to optimize the remaining code and have opened up a question on stack concerning the algorithm. I'm hoping the wizards over there might be able to suggest some ways to make things calculate faster.

https://stackoverflow.com/questions/55249029/exploding-keys-and-values-in-a-dictionary-subset-in-python

Zeitsperre:
@huard 

I'm still waiting to see if there are any persons interested in optimizing our upstream basin identification algorithm but in the meantime, I looked into leveraging GeoPandas to speed up the identification of the basin.

It turns out 1) using GeoPandas.from_features() is about much slower than using Fiona (13sec vs 4sec) and 2) using GeoPandas.from_file() is still slower (11sec vs 4sec). 
```
import datetime
import time

now = time.time()
1)
# with fiona.Collection(vector_file, 'r', crs=shape_crs) as src:
#     gdf = gpd.GeoDataFrame.from_features(src, crs=shape_crs)
#     print(gdf[gdf.contains(location)])
2)
# gdf = gpd.GeoDataFrame.from_file(vector_file, crs=shape_crs)
#     print(gdf[gdf.contains(location)])
then = time.time() - now
raise UserWarning(str(datetime.timedelta(seconds=int(then))))
```

My theory is that we lose a lot of time when unpacking the zip file and running the upstream identification algorithm. To speed this up I'm thinking that if the data exists on a GeoServer, we could leverage the WFS capabilities to have the server return the feature of interest based on the points which would remove the need for expanding zip files.

To address the upstream identification problem, we could go a bit further in pre-calculating the network structure and storing the main basins and levels as python pickles. I imagine that reading a pickle is quicker than building a tree, but I can't say with certainty or authority. What do you think?

huard:
So we have two issues: 
- file access speed
- the algorithm 
How long is it to only read one column from the database ? If it's substantial, then yes, maybe let's start considering storing that as a standalone file on GeoServer. 

The next issue is the algorithm. I come back to the idea that the Pfafstetter number let's us determine with one operation whether the basin is upstream or not. It's not easy to parse, agreed, but that seems like the most efficient way to get what we want.

Zeitsperre:
You're right about the Pfafstetter approach. Something to look into once I have time in the next two week. Posting this until then: http://proceedings.esri.com/library/userconf/proc01/professional/papers/pap1008/p1008.htm.

huard:
@Zeitsperre Have you measured a speed up using the tables ?

PR #95, opened 2019-03-26
Added global netcdf attributes 

huard:
## Overview

Adds global netcdf attributes to output files, include the `model_id` attribute. 

Note that comments with commas or spaces are chopped, so only the first word is saved. Not a big issue for now, but something to consider.


richardarsenault:
Tested, works, merging.

PR #96, opened 2019-03-27
Refactored delineation process

huard:
## Overview

Refactored delineation process. There were a few bugs with the output data handling. 

@Zeitsperre Use
```
response.outputs['name'].file = <path to file created by process>
```
or
```
response.outputs['name'].data = <string or bytes created by process>
```

Changes:

* Changed the signature a bit
* Location is now "lat, lon" instead of "(lat, lon)"
* Actual computations are done by functions inside utilities/gis.py


huard:
@Zeitsperre The upstream collection returns a feature collection of one item. Is it safe to return the fist feature instead ? Otherwise the behavior changes when `aggregate_upstream` is True.

richardarsenault:
@huard that is a good question, I had a few problems with the FeatureCollection objects which limits the use of "collect_upstream". Having it return a feature all the time would be better.

huard:
@Zeitsperre Also, by calling `up.dissolve(by='MAIN_BAS')`, by the default the aggregation process uses the `first` method, meaning that the properties of the output are the properties of the downstream feature, not the properties of the aggregated features. The result is that the properties don't really match anymore. We should either scrap properties that make no sense anymore of compute them correctly (for example by summing areas).

huard:
I got about a 4x speedup on the test compared to the original version.

PR #99, opened 2019-03-27
Simplied zonal stats and fixed bugs in response

huard:
## Overview

Refactoring of zonal stats process.

### DIFF
--- a/tests/test_wps_generic_zonal_stats.py
+++ b/tests/test_wps_generic_zonal_stats.py
@@ -37,13 +37,8 @@ class TestGenericZonalStatsProcess:
         assert_response_success(resp)
         out = get_output(resp.xml)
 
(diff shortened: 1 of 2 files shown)


tests/test_wps_generic_zonal_stats.py
```diff
@@ -109,15 +100,10 @@ def test_simple_geojson(self):
 
         assert_response_success(resp)
         out = get_output(resp.xml)
+        feature = json.loads(out['statistics'])['features'][0]
```
Zeitsperre: This only checks the first feature, right? Why not all?

huard: It's the same logic as the original test. 
And it's just a smoke test, we're still not checking the actual values.

PR #101, opened 2019-03-28
lake must be True with aggregate_upstream

huard:
## Overview

Fix failure when lake is False and aggregate is True.

### DIFF
--- a/raven/processes/wps_hydrobasins_shape_selection.py
+++ b/raven/processes/wps_hydrobasins_shape_selection.py
@@ -110,7 +110,8 @@ class ShapeSelectionProcess(Process):
             # Find feature containing location
             feat = next(src.filter(bbox=bbox))
             hybas_id = feat['properties']['HYBAS_ID']
-
+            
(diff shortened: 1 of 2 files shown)


raven/processes/wps_hydrobasins_shape_selection.py
```diff
@@ -95,12 +95,6 @@ def _handler(self, request, response):
...
         table = DATA / 'hybas_{}na_lev{:02}.csv'.format('lake_' if lakes else '', level)
-
-        if shape_description != 'hydrobasins_lake_na_lev12':
-            msg = 'Options for HydroBASINS levels and lakes will be in prod'
-            LOGGER.error(NotImplementedError(msg))
-            shape_description = 'hydrobasins_lake_na_lev12'
-
         shape_url = TESTDATA[shape_description]
 
         # Why not just map(float, lonlat.split(',')) ?
```
Zeitsperre: I had completely forgotten about map(). I hadn't thought of it!

PR #112, opened 2019-04-16
Fix 109

### LINKED ISSUE #109: Need to define classes for land-cover aggregation from Tiff
## Description

Need to go from ~20 classes down to ~5 for regionalization.

Zeitsperre:
## Overview

This PR fixes #109 

Changes:

* Added a specific zonal stat process using the CEC NALCMS (2010) groupings suggested in #109 

## Additional Information

### DIFF
--- a/raven/processes/wps_nalcms_zonal_stats.py
+++ b/raven/processes/wps_nalcms_zonal_stats.py
@@ -1,5 +1,6 @@
 import logging
 import json
+import tempfile
 
 from pywps import LiteralInput, ComplexInput
 from pywps import ComplexOutput
@@ -7,7 +8,7 @@ from pywps import Process, FORMATS
 from pywps.app.Common import Metadata
 from rasterstats import zonal_stats
 
-from raven.utils import archive_sniffer, crs_sniffer, single_file_check
+from raven.utils import archive_sniffer, crs_sniffer, generic_vector_reproject, single_file_check
 
 LOGGER = logging.getLogger("PYWPS")
 
@@ -98,9 +99,15 @@ class NALCMSZonalStatisticsProcess(Process):
 
         vec_crs, ras_crs = crs_sniffer(vector_file), crs_sniffer(raster_file)
 
-        if ras_crs != vec_crs:
-            msg = 'CRS for files {} and {} are not the same.'.format(vector_file, raster_file)
+        if vec_crs != ras_crs:
+            msg = 'CRS for files {} and {} are not the same. Reprojecting...'.format(vector_file, raster_file)
             LOGGER.warning(msg)
+
+            # Reproject full vector to preserve feature attributes
+            projected = tempfile.NamedTemporaryFile(prefix='reprojected_', suffix='.json', delete=False,
+                                                    dir=self.workdir).name
+            generic_vector_reproject(vector_file, projected, driver='GeoJSON', source_crs=vec_crs, target_crs=ras_crs)
+            vector_file = projected
 
         try:
             stats = zonal_stats(
--- a/raven/processes/wps_nalcms_zonal_stats.py
+++ b/raven/processes/wps_nalcms_zonal_stats.py
@@ -12,27 +12,18 @@ from raven.utils import archive_sniffer, crs_sniffer, generic_vector_reproject, 
 
 LOGGER = logging.getLogger("PYWPS")
 
-NALCMS_CATEGORIES = {
-    1: 'Forest',
-    2: 'Forest',
-    3: 'Forest',
-    4: 'Forest',
-    5: 'Forest',
-    6: 'Forest',
-    7: 'Shrubs',
-    8: 'Shrubs',
-    9: 'Grass',
-    10: 'Grass',
-    11: 'Shrubs',
-    12: 'Grass',
-    13: 'Grass',
-    14: 'Wetland',
-    15: 'Crops',
-    16: 'Grass',
-    17: 'Urban',
-    18: 'Water',
-    19: 'SnowIce',
+categories = {
+    'Forest': [1, 2, 3, 4, 5, 6],
+    'Shrubs': [7, 8, 11],
+    'Grass': [9, 10, 12, 13, 16],
+    'Wetland': [14],
+    'Crops': [15],
+    'Urban': [17],
+    'Water': [18],
+    'SnowIce': [19]
 }
+
+NALCMS_CATEGORIES = {i: cat for (cat, ids) in categories.items() for i in ids}
 
 
 class NALCMSZonalStatisticsProcess(Process):
@@ -94,6 +85,9 @@ class NALCMSZonalStatisticsProcess(Process):
 
         vectors = ['.gml', '.shp', '.gpkg', '.geojson', '.json']
         vector_file = single_file_check(archive_sniffer(shape_url, working_dir=self.workdir, extensions=vectors))
+
+
+        # TODO: Set up the WCS call using the gis.get_bbox and gis.get_dem functions
         rasters = ['.tiff', '.tif']
         raster_file = single_file_check(archive_sniffer(raster_url, working_dir=self.workdir, extensions=rasters))
 
--- a/raven/processes/wps_nalcms_zonal_stats.py
+++ b/raven/processes/wps_nalcms_zonal_stats.py
@@ -13,7 +13,29 @@ from raven.utils import archive_sniffer, crs_sniffer, generic_vector_reproject, 
 
 LOGGER = logging.getLogger("PYWPS")
 
-categories = {
+TRUE_CATEGORIES = {
+    1: 'Temperate or sub-polar needleleaf forest',
+    2: 'Sub-polar taiga needleleaf forest',
+    3: 'Tropical or sub-tropical broadleaf evergreen forest',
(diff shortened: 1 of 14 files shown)


Zeitsperre:
@huard The CEC NALCMS land classification data set uses a non-standard proj4 definition so the chances of a vector file sharing a CRS with this layer are near zero. As such, I needed to add a more robust vector reprojection (not just for individual geometries) method to `utils.py`. I'm also testing to see if the `GeoJSON` driver with fiona works as it could make file-handling a bit easier.

I'll configure the tests tomorrow and see if it the classifications work. This should work with everything, inlcuding GeoPackages.

huard:
There seem to be a related failing test. Ideas ?

richardarsenault:
@huard @Zeitsperre I just tested and the failed test is due to the assertion check line 39 in "test_wps_nalcms_zonal_stats.py". It seems we are checking to see if the individual values (from 135:228) are a subset of the stats variable, whereas the stats are a dict of statistics (mean, min, max, count...). Not sure what the goal is here but that seems to be the problem.

Ideally this code should return the percentages of each class, i.e. 35% grass, 45% forest, 10% water, etc.

Zeitsperre:
@richardarsenault @huard 

Sorry for the mix-up but this branch is still missing a few things that I'm addressing this week. The tests are not set up properly and there are a few particularities/problems that still need addressing (such as how to return the true pixelsounts and their values for a WCS-served layer).

Zeitsperre:
@richardarsenault 

Function seems to be working on my end. Some cleaning up to do but it looks like this is now outputting what we want.

huard:
Yes, let's wait. When do you think it will be done ?

raven/processes/wps_nalcms_zonal_stats.py
```diff
@@ -0,0 +1,134 @@
...
+from pywps import ComplexOutput
+from pywps import Process, FORMATS
+from pywps.app.Common import Metadata
+from rasterstats import zonal_stats
+
+from raven.utils import archive_sniffer, crs_sniffer, generic_vector_reproject, single_file_check
+
+LOGGER = logging.getLogger("PYWPS")
+
+NALCMS_CATEGORIES = {
```
huard: This is just style, but I think it would be cleaner to do 
```
categories = {'Forest':[1,2,3,4,5,6], 'Shrubs':[7,8,11], ...} 
NALCMS_CATEGORIES = {i:cat for (cat, ids) in categories.items() for i in ids}
```

raven/processes/wps_nalcms_zonal_stats.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def __init__(self):
+        inputs = [
+            ComplexInput('shape', 'Vector Shape',
+                         abstract='An ESRI Shapefile, GML, JSON, GeoJSON, or single layer GeoPackage.'
+                                  ' The ESRI Shapefile must be zipped and contain the .shp, .shx, and .dbf.'
+                                  ' The shape and raster should have a matching CRS.',
+                         min_occurs=1, max_occurs=1,
+                         supported_formats=[FORMATS.GEOJSON, FORMATS.GML, FORMATS.JSON, FORMATS.SHP]),
+            ComplexInput('raster', 'Gridded Land Use raster data set',
```
huard: Once the DEM is on GeoServer, this will not be necessary correct ?

raven/processes/wps_nalcms_zonal_stats.py
```diff
@@ -0,0 +1,134 @@
...
+            stats = zonal_stats(
+                vector_file, raster_file, stats=['count', 'nodata'],
+                band=1, categorical=True, category_map=NALCMS_CATEGORIES, all_touched=touches,
+                geojson_out=geojson_out, raster_out=False)
+
+            if not geojson_out:
+                response.outputs['statistics'].data = json.dumps(stats)
+
+            else:
+                if len(stats) > 1:
```
huard: It's really a pain on the client side when the outputs change their topology depending on input parameters. In other words, we expect that outputs are a function of the inputs, but the container of the outputs should not change. Here what does the FeatureCollection add ? I suggest outputs are either always FeatureCollections (with one item sometimes) or never FeatureCollections.

tests/test_wps_nalcms_zonal_stats.py
```diff
@@ -0,0 +1,71 @@
...
+        )
+
+        resp = client.get(
+            service='WPS', request='Execute', version='1.0.0', identifier='nalcms-zonal-stats', datainputs=datainputs)
+
+        assert_response_success(resp)
+        out = get_output(resp.xml)
+
+        stats = json.loads(out['statistics'])[0]
+        print(stats)
```
huard: remove print

Zeitsperre:
@huard Once @moulab88 resets the GeoServer instance, I can set it up and close this.

raven/processes/wps_nalcms_zonal_stats.py
Zeitsperre: Interesting. Sure.

raven/processes/wps_nalcms_zonal_stats.py
Zeitsperre: FeatureCollections are the GeoJSON standard for multi-feature vector files. I can set things so that they always output FeatureCollections regardless of number of features and there shouldn't be any problems.

raven/processes/wps_nalcms_zonal_stats.py
Zeitsperre: Correct. It would be interesting to provide different years [2000, 2010, 2020(?)] for the NALCMS data set but they would all be served via WCS provided by the PAVICS instance.

richardarsenault:
@Zeitsperre Hi, any news on this front? As soon as it is merged I will start batch-processing catchment property extraction. Thanks!

Zeitsperre:
@moulab88 is back. I'll be asking him to restart our instance and see if I can get everything configured hopefully by tomorrow.

Zeitsperre:
@huard 

This was a bit of a headache but (smoke) tests are passing and manual inspection suggests that values are accurate. The NALCMS 2010 data is using a generic LAEA projection and because of the size of the file and my reservations about reprojecting land use rasters, I've opted to demand bounding boxes using those coordinates. As WCS 2.0.1 uses different axis names for projected data (easting and northing), I added a `get_nalcms_wcs` function that formats the call using those axes. This could probably be collapsed into a `get_raster_wcs` function with `projected` or `geographic` as the wcs axis type keywords.

@richardarsenault 

The data is now linked to the local GeoServer imagery and will default to it if no land use raster is provided in `nalcms-zonal-stats` calls. Have fun!

raven/processes/wps_nalcms_zonal_stats.py
```diff
@@ -0,0 +1,144 @@
...
+            stats = zonal_stats(
+                vector_file, raster_file, stats=['count', 'nodata'],
+                band=band, categorical=True, category_map=NALCMS_CATEGORIES, all_touched=touches,
+                geojson_out=geojson_out, raster_out=False)
+
+            if not geojson_out:
+                response.outputs['statistics'].data = json.dumps(stats)
+
+            else:
+                feature_collect = {'type': 'FeatureCollection', 'features': stats}
```
huard: Am I correct thinking the output's type depends on `geojson_out` ? As discussed earlier, that's not ideal.

Zeitsperre: Ah yes, this was changed in the other branch. I'll update this here as well.

Zeitsperre: Would it make sense to split the functions based on the desired output? If hydro models don't need geographic information, they can use `wps_simple_zonal_stat` and for functions that do, use `wps_geojson_zonal_stat`?

huard: Not clear to me what would be best. What are the typical use cases ? And also, I'd like to avoid code duplication, and to do that, we'd need a base class. So let's keep it simple for now, and think about improvements for later.

Zeitsperre: I would tend towards preserving the GeoJSON format since that provides all the same information and preserves the geometry, but I'm biased. Also, saving this information within a `FeatureCollection` provides an easier means of loading and parsing information for each feature.

huard: Ok, I just don't want the output's type to change based on arguments. Let's return a FeatureCollection all the time and users will have to dig inside it to get the stats. Or split these two outputs entirely, whatever sounds most convenient.

Zeitsperre: Last question before I push changes: Can we specify in the `wps_nalcms_zonal_stats.py` to output either the simplified values or the real values for pixels? I can see having the real categories being useful for other applications.

huard: Yes, makes sense.

Zeitsperre:
Lots of refactoring and revisited some of the older GIS processes to fix metadata and add the helper functions we've created recently. All these processes (except the hydrobasin-based process) are now connected to the WCS service. GeoJSON is now the standard output for most shape-based stats functions. 

To makes things simpler, the `wps_generic_terrain_analysis` now requires a vector. I hope this isn't an issue.

PR #138, opened 2019-06-05
Added complete watershed delineation and property extraction notebook

richardarsenault:
This is a notebook that uses a lat/long of an outlet as input, and computes:
1- Watershed boundaries
2- NALCMS land use
3- Shape properties
4- Terrain analysis

All using the Geoserver instead of the predefined DEMs, etc.

Only remaining point is how to get the info from the notebook to the user, i.e. how to pass a boundary shapefile or json?


huard:
I fixed #140 and #139 . 

You should be able to pass the path to a shapefile or geojson to the server.

PR #144, opened 2019-06-09
Updated Regionalization data to correct NALCMS

richardarsenault:
Also uploading the code used to extract the data from the 5797 basins. Will need to be rerun on a more powerful server to extract the data for the remaining huge catchments (memory problems on my side).


tests/testdata/regionalisation_data/GetResgionalizationData.py
```diff
@@ -0,0 +1,164 @@
...
+url = "http://localhost:9099/wps"
+wps = WPSClient(url, progress=True)
+
+    
+nbBasin=5797
+
+for i in range(1, nbBasin+1):
+
+    #Read the current dataset
+    tmp=pd.read_csv('gauged_catchment_properties.csv')
```
huard: This function reads and writes from the same file. That seems weird.

huard: And why do you need to read it at each iteration ?

tests/testdata/regionalisation_data/GetResgionalizationData.py
```diff
@@ -0,0 +1,164 @@
...
+#import os
+import time
+
+url = "http://localhost:9099/wps"
+wps = WPSClient(url, progress=True)
+
+    
+nbBasin=5797
+
+for i in range(1, nbBasin+1):
```
huard: matlab habits die hard ?

tests/testdata/regionalisation_data/GetResgionalizationData.py
```diff
@@ -0,0 +1,164 @@
...
+                print('Doing Basin: ' + str(i))
+                shape2='/home/ets/SHP_ZIP/Basin_' + str(i) +'.zip'         
+                
+                resp = wps.shape_properties(shape=shape2, projected_crs=32198)
+                
+                while resp.isNotComplete():
+                    time.sleep(1)
+                
+                [properties]=resp.get(asobj=True)
+                Area[i-1]=properties[0]['area']/1000000.0
```
huard: I suggest we change the returned values so you can use the properties dictionary directly, without transformation.

richardarsenault:
I see the confusion, sorry! I added that code as a backup so we can regenerate the files if needed, but in theory this will not be a production code. The review was mostly for the database. I built it that way because when the raven server crashes on my small laptop, I lose everything if it's not written. So I structured it in a way that I update the file on each iteration, and I can then redo a pass for those that were not successful. 

I don't think we will use it for anything else really, it was hardcoded for those 5797 catchments. I expected to run it one last time on the PAVICS server when it will be up for the larger catchments. What do you think?

tests/testdata/regionalisation_data/GetResgionalizationData.py
richardarsenault: I had to preallocate my values first so it would work with the "RunSuccess" Flag. Not sure it's useful anymore. And yes, I still do 95% of my coding in Matlab 😄

huard:
Don't change it if it works, but I think it could be useful to have a clean version of this code as an example of how to build a DB for regionalization.

Zeitsperre:
@richardarsenault If you haven't started yet. I can do a first-pass cleaning of this later today or tomorrow morning.

huard:
@Zeitsperre  Please focus on the averaging notebook.

PR #147, opened 2019-06-13
Enabled multipolygon_check and minor bugfix

Zeitsperre:
## Overview

Changes:

* Added a multipolygon check 

## Related Issue / Discussion

I've noticed there are instances where a multipolygon can create problems. All this check does is log that the features provided are multipolygons.

## Additional Information


Zeitsperre:
This creates more problems than it addresses. Closing.

PR #149, opened 2019-06-13
Fix 132

### LINKED ISSUE #132: Example of meteo netCDF subset for watershed.
## Description

Create a notebook showing how to subset and aggregate climate variables for a watershed and feed the output to raven.

Zeitsperre:
## Overview

This PR fixes #132 

Changes:

* Added a script that attempt to run the flyingpigeon (upstream/master) process `subset-wfs-polygon`.

## Related Issue / Discussion
#132 

## Additional Information


huard:
Status ?

Zeitsperre:
Blocked. Still waiting to hear from @cehbrecht or @nilshempelmann about this in https://github.com/bird-house/flyingpigeon/issues/287

richardarsenault:
Is there another way we can get meteorological/climate data from a polygon using what we already have? So we can show that the hydro part can communicate with the PAVICS server. It can be a convoluted way for now, so we can at least show that capability during the demo.

huard:
Looking into this.

richardarsenault:
Is there any development on this side lately? We will need this component to allow users to do get climate data for their models. Thanks.

nilshempelmann:
@Zeitsperre sorry for beeing blocked, bird-house/flyingpigeon#287 is related to an oranos development `subset_wfs_polygon`. It is more senceful you are fixing that within the ouranos team. 
@richardarsenault: based on predefined polygons you can use the processes in flyingpigeon https://flyingpigeon.readthedocs.io/en/latest/processes_api.html#subset-processes-api 

Let me know if I can help somewhere

Zeitsperre:
If we can integrate the changes from @huard FP fork, the issues are resolved and work can continue on this PR when I get another chance at it.

nilshempelmann:
@zeitsperre you need the fix integrated in FP master? Make a PR.

Zeitsperre:
@huard can you open a PR with your changes to FP to upstream master?

huard:
@Zeitsperre Hasn't it been merged already ? https://github.com/bird-house/flyingpigeon/pull/288

Zeitsperre:
It looks like it has. I didn't catch that change in my review of the past two weeks. I'll continue from here. Thanks!

raven/processes/wps_generic_shape_properties.py
```diff
@@ -75,6 +75,8 @@ def _handler(self, request, response):
                     for feature in src:
                         geom = shape(feature['geometry'])
 
+                        multipolygon_check(geom)
```
huard: Is this purely for debugging purpose ? It only logs information.

raven/processes/wps_hydrobasins_shape_selection.py
```diff
@@ -136,10 +139,10 @@ def _handler(self, request, response):
 
                 feat = json.loads(agg.to_json())['features'][0]
                 response.outputs['feature'].data = json.dumps(feat)
-                response.outputs['upstream_ids'].data = json.dumps(up['HYBAS_ID'].tolist())
+                response.outputs['upstream_ids'].data = json.dumps(up['id'].tolist())
```
huard: I guess these are the IDs needed to extract WFS features afterward ? Won't this break code using specifically the HYBAS_ID ?

huard: Maybe there is no such code, but this is the kind of things that should be clearly highlighted in the PR description.

raven/processes/wps_generic_shape_properties.py
Zeitsperre: At the moment, yes. It's written for cases where the algorithm can only handle a polygon and not a multipolygon. Some of the processes we have are sensitive to this distinction so this idea is to log this information so that the user understands why something has failed.

raven/processes/wps_hydrobasins_shape_selection.py
Zeitsperre: Right. Yes, I'll be sure to be clear about changes like this in the future.

As it stands, there is no code that makes use of the `HYBAS_ID`. Since the database rests solely accessible via WFS, the data is more easily queried using the GeoServer address and this `id` (when written to file is called the `gml_id`). I felt the change was necessary as the `upstream_ids` were not the `HYBAS_ID`'s either, instead being the WFS `id`'s. Now everything is consistent.

Zeitsperre:
> I think the delineation workflow will need to be updated after the change of name for the process.

I'll get to that right now. and merge after that change.

PR #153, opened 2019-06-18
Regionalization notebook

richardarsenault:
Notebook for regionalization. Works except for 2 things:

1-WPS hangs and will not exit infinite loop until you hit enter at the command line in the terminal that is running RAVEN.  The terminal will stop searching for "something" and return the response.

2- The plotting of the "ensemble" hydrographs is returning a bitmap image instead of a line graph.


huard:
1. Fixed. It was the version attribute. We launch raven with no parameters to get a string with the version. But I think there was a change with the latest version and now it expects a Return. 
@julemai I've been able to fix this, but why would raven wait for return ?

huard:
For 2 you need to specify what kind of plot you want when it's not obvious (multiple dimensions).

PR #164, opened 2019-07-05
Notebooks connect to PAVICS

### LINKED ISSUE #163: Use pavics.ouranos.ca as the server for notebooks
## Description

- Change localhost to pavics.ouranos.ca
- Check that the notebooks still work

huard:
## Overview

This PR fixes #163

MOHYSE Ostrich fails for some reason.


huard:
Richard, you can deploy to the server by running
```
bumpversion patch
git push
git push --tags
```

tlvu:
Jenkins run log: http://jenkins.ouranos.ca/job/PAVICS-e2e-workflow-tests/job/enable-raven-notebooks-in-jenkins/6/console

A few errors found:

Jenkins didn't know to wait for the background process to complete:
```
ProcessIsNotComplete: Please wait ...

```

Jupyter env missing a module, this must have been added recently to the notebooks, I'll take care of this.
```
ModuleNotFoundError: No module named 'rasterio'
```

Bugs in birdy?
```
18:13:32  /usr/local/envs/birdy/lib/python3.6/site-packages/birdy/client/converters.py in convert(output, path, converters, verify)
18:13:32      271             pass
18:13:32      272 
18:13:32  --> 273     warnings.warn(UserWarning("No converter was found for {}".format(output.identifier)))
18:13:32      274     return output.reference
18:13:32      275 
18:13:32  
18:13:32  AttributeError: 'str' object has no attribute 'identifier'

18:13:32  /usr/local/envs/birdy/lib/python3.6/site-packages/birdy/client/converters.py in convert(self)
18:13:32      156         try:
18:13:32      157             # try OpenDAP url
18:13:32  --> 158             return xr.open_dataset(self.url)
18:13:32      159         except IOError:
18:13:32      160             # download the file
18:13:32  
18:13:32  AttributeError: 'XarrayConverter' object has no attribute 'url'
```

And a bunch of textual output different we need to regex match to escape them, one example:
```
[2019-07-05T22:13:32.758Z]   - 19-07-05 21:03:46 by Rave
[2019-07-05T22:13:32.758Z]   ?             ^^^^
[2019-07-05T22:13:32.758Z]   + 19-07-05 22:10:16 by Rave
[2019-07-05T22:13:32.758Z]   ?           ++ + ^
```

So there are still some work before these notebooks can pass on Jenkins.

If these notebooks pass manually when running by hand it's good for now.  We'll fix them for Jenkins later.

huard:
You'll need the latest birdy. I updated it this week.

tlvu:
Latest released birdy is still birdhouse-birdy=0.6.2 on conda.

richardarsenault:
I had to do the same thing, and there is a branch on the birdy git that is 0.6.3. It worked, but not sure we want to go there for the deployed version.

tlvu:
Same here, I'd rather wait for an official release.  I don't think anyone is waiting after this fix.  I was just re-building the jupyter server image to include the missing rasterio package and I thought I would pickup a newer birdy.

huard:
@Zeitsperre Do you have time to check if this is working now with the new birdy version ?

Zeitsperre:
@huard 

Many of the calls have been updated in recent PRs and it's impossible to test against machines when their APIs have changed. As such, I've changed all the links to localhost for the time being so that Travis CI can run the tests off its local Raven service. This will ultimately be changed back to PAVICS once the notebooks have been cleaned and are passing. I need to tag and push a new version of RAVEN so that the notebooks can be properly tested on the PAVICS stack. 

Also, similar to to the testing approach to notebooks present in pavics-sdi, I'm trying to run `pytest --nbval` with what we have available here but I'm running into an issue with the progress bar output that some notebooks are using. I'm about 90% certain that the progress bar is treated as a final output by pytest, causing it to run through tests faster than Raven can serve reponse objects. Rather than disable this feature (since it's pretty neat to show to users) I'm looking for a way to tell pytest to wait until the progress bar is done. Do you know of anything?

huard:
No. One easy way out is to skip the notebooks with a progress bar in pytest, and run them manually.

tlvu:
Well the whole point is we automatically test all the notebooks, otherwise, how do we know, *for sure*, the notebooks we give to the user are actually working?  Let's take a bit more time on this than immediately remove this notebook from the automated tests.

Is there a sort of status url we can somehow "ping" so we can loop on it before executing the next step?

Or, at the very least, if the progress bar feature is not so important to test and not so important to demo to user, we can simply not use it and we get to keep the notebook under automated tests.  We can add a comment that the progress bar option is available if user want to try it manually.

Zeitsperre:
Update: Things seem to be coming along!

Screenshot from 2019-09-05 13-07-47

@huard I'm wondering if the OSTRICH model notebooks are set up to always return different values. I've tried manually running them a few times and I seem to ALWAYS return different values. I'm hoping this is the case or else I might have to consult a (medical) doctor.

For some of the notebooks, errors are thrown because of differences in the internals of the response's "help" text? 

Screenshot from 2019-09-05 13-15-07

I've elected to disable evaluation of these cells since they're just statements anyway.

huard:
You're not mad. Ostrich uses random number seeds. You'll get a consistent answer only if those seeds are identical one run from the next. Ostrich should converge if the number of iterations is large enough, but this would make the tests take far too long. 

In principle, setting `random_seed=0` in the call to ostrich should yield the same answer from run to run. Good catch.

Zeitsperre:
I've added another makefile recipe for running the notebooks (`make test_nb`). I might look into porting this to other birds that are running notebooks. It seems to work fine.

The majority of failures can be addressed by sanitizing a few more regex patterns (e.g. warning messages sent from libraries under my user "/home/tjs" != "/home/travis") and rerunning some Raven notebooks in a fresh local environment (ie: `watershed_subset.ipynb` lists versions at the start; `Delineation_workflow.ipynb` reports back data with different significant digits). The most recent regex changes have been ported to @tlvu 's raven branch in [PAVICS-e2e-workflow-tests](https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/tree/enable-raven-notebooks-in-jenkins) but there will be more to add.

Screenshot from 2019-09-05 16-52-11

Only 13 failures to go!

huard:
@tlvu @Zeitsperre  

If we want to test notebooks locally (or on travis-ci), we want birdy to hit the localhost server. However, we'd like the notebooks on jupyterlab and in the documentation to hit the official pavics server. 

So far, we've been changing the url back and forth directly in the notebooks. That is, setting the url to localhost when testing. When it works, we change it back to pavics and then push the change and build the docs. It works, but it's a bit of a pain. 

Should we set the URL to an environment variable and not include it explicitly in the notebooks ? 

Suggestions ?

tlvu:
> Should we set the URL to an environment variable and not include it explicitly in the notebooks ?

Could use the same trick used here https://github.com/Ouranosinc/pavics-sdi/blob/d9d72e299e0f41ca21a76ceb882fbb6da16d9faa/docs/source/notebooks/esgf-dap.ipynb#L83

So something like 
```
url = os.environ.get('PYWPS_URL', "https://pavics.ouranos.ca/twitcher/ows/proxy/raven/wps")
```
and when running locally or via Travis-CI, `export PYWPS_URL="http://localhost:port/wps"` in the shell before starting jupyter local server to run the notebook.

However, if something like this `output_netcdf='https://pavics.ouranos.ca/wpsoutputs/1bb76e5e-827a-11e9-be8a-0242ac19000f/out.nc'` is displayed, we won't be able to replace it for the localhost version.

huard:
Ok, will try this.

Zeitsperre:
@huard A new version is being deployed now to the test server. Once we have that. we can start ensuring that notebook URLs are functioning.

Zeitsperre:
@huard The newest RAVEN build is failing at the DescribeProcess() call. This is very strange given that all tests are passing when running locally. This is the trace when simply trying to instantiate a WPSClient with any URL to raven (either using localhost or https):
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-bb68a39db724> in <module>
      6 url = "http://localhost:9099/wps"
      7 url = "https://pavics.ouranos.ca/twitcher/ows/proxy/raven/wps"
----> 8 wps = WPSClient(url, progress=True)

/usr/local/envs/birdy/lib/python3.6/site-packages/birdy/client/base.py in __init__(self, url, processes, converters, username, password, headers, auth, verify, cert, verbose, progress, version)
    112             raise
    113 
--> 114         self._processes = self._get_process_description(processes)
    115 
    116         # Build the methods

/usr/local/envs/birdy/lib/python3.6/site-packages/birdy/client/base.py in _get_process_description(self, processes)
    144             if owslib.__version__ > '0.17.0':
    145                 # Get the description for all processes in one request.
--> 146                 ps = self._wps.describeprocess('all')
    147                 return OrderedDict((p.identifier, p) for p in ps)
    148             else:

/usr/local/envs/birdy/lib/python3.6/site-packages/owslib/wps.py in describeprocess(self, identifier, xml)
    298 
    299         # build metadata objects
--> 300         processes = self._parseProcessMetadata(rootElement)
    301 
    302         if identifier == 'all':

/usr/local/envs/birdy/lib/python3.6/site-packages/owslib/wps.py in _parseProcessMetadata(self, rootElement)
    369         processes = []
    370         for processDescriptionElement in processDescriptionElements:
--> 371             process = Process(processDescriptionElement, verbose=self.verbose)
    372 
    373             # override existing processes in object metadata, if existing already

/usr/local/envs/birdy/lib/python3.6/site-packages/owslib/wps.py in __init__(self, elem, verbose)
   1562         self.dataInputs = []
   1563         for inputElement in elem.findall('DataInputs/Input'):
-> 1564             self.dataInputs.append(Input(inputElement))
   1565             if self.verbose is True:
   1566                 dump(self.dataInputs[-1], prefix='\tInput: ')

/usr/local/envs/birdy/lib/python3.6/site-packages/owslib/wps.py in __init__(self, inputElement)
   1264 
   1265         # <LiteralData>
-> 1266         self._parseLiteralData(inputElement, 'LiteralData')
   1267 
   1268         # <ComplexData>

/usr/local/envs/birdy/lib/python3.6/site-packages/owslib/wps.py in _parseLiteralData(self, element, literalElementName)
   1119                 if sub_element.tag.endswith('DefaultValue'):
   1120                     self.defaultValue = getTypedValue(
-> 1121                         self.dataType, sub_element.text)
   1122 
   1123                 if sub_element.tag.endswith('AllowedValues'):

/usr/local/envs/birdy/lib/python3.6/site-packages/owslib/util.py in getTypedValue(data_type, value)
    578         return int(value)
    579     elif data_type == 'float':
--> 580         return float(value)
    581     elif data_type == 'string':
    582         return str(value)

TypeError: float() argument must be a string or a number, not 'NoneType'
```

tlvu:
@Zeitsperre attached are the conda env export of raven 0.7.0 docker image and the latest 0.8.1-beta that seems to have problem.  Hope it helps.

[raven-0.7.0.conda-env.txt](https://github.com/Ouranosinc/raven/files/3764694/raven-0.7.0.conda-env.txt)

[raven-0.8.1-beta.conda-env.txt](https://github.com/Ouranosinc/raven/files/3764695/raven-0.8.1-beta.conda-env.txt)

Also raven code diff between those 2 versions: https://github.com/Ouranosinc/raven/compare/v0.7.0...v0.8.1-beta

tlvu:
@cehbrecht quick question, the conda birdhouse channel used to publish `owslib=0.17.1b4`, is whatever was in that beta release part of the upstream `owslib=0.18.0`?

We seems to have an issue with `owslib` between Raven and birdy client.

huard:
Got it (I think) . Could be a PyWPS change. The template for literal data is different between 4.2.1 and 4.2.2: 

See (4.2.2)
https://github.com/geopython/pywps/blob/313a3b456b351edb3e76097c2bb0273f25ed8d46/pywps/templates/1.0.0/describe/literal.xml#L35

and (4.2.1)
https://github.com/geopython/pywps/blob/3383f65ab30f17505e632492c00780ca5ab20114/pywps/templates/1.0.0/describe/literal.xml#L33

With 4.2.2, the response includes a DefaultValue even if it should not. ```put.data``` is empty, but the ```if put.data is not none``` seems to be true. Should it be something like ```if not none(put.data)``` ?

huard:
@davidcaron This seems to be a change you introduced a while back. Are you positive that the syntax to check for none is right ?

davidcaron:
> 
> 
> @davidcaron This seems to be a change you introduced a while back. Are you positive that the syntax to check for none is right ?

I had to test to make sure:

```python
from jinja2 import Template


output = Template(
    "{% if variable is not none %}if variable is not none{% endif %}"
).render(variable=None)

assert output == ""

output = Template(
    "{% if variable is not none %}if variable is not none{% endif %}"
).render(variable="something")

assert output == "if variable is not none"

error_raised = False
try:
    output = Template("{% if none(variable) %}if none(variable){% endif %}").render(
        variable=None
    )
except TypeError:
    error_raised = True

assert error_raised
```

huard:
Thanks for testing and the explanation. So I suppose put.data is not really None when it should be.

huard:
So what happens is that put.data is just not defined, and the ```is none``` clause evaluates to False. 

I suggest replacing ```{% if variable is not none %}``` by ```{% if variable is defined %}```.

davidcaron:
> So what happens is that put.data is just not defined, and the `is none` clause evaluates to False.
> 
> I suggest replacing `{% if variable is not none %}` by `{% if variable is defined %}`.

I think it should be `{% if variable is defined and variable is not none %}`

huard:
https://github.com/geopython/pywps/pull/499

huard:
So the test_nb make command sets environment variables RAVEN_WPS_URL and FLYINGPIGEON_WPS_URL before launching pytest --nbval 

Each notebook then sets the WPS url either to the environment variable, or if not defined, the pavics url. This means that we can run the notebook test suite locally, on travis, or on jupyterlab (where these environment variables should not be defined).

PR #165, opened 2019-07-25
Fix 140

### LINKED ISSUE #140: bug in nalcms category aggregation
## Description

zonal_stats doesn't sum the contribution of each category, hence the mismatch between the pixel count and count for individual categories. Fixed. Will issue PR with delineation notebook.

Zeitsperre:
## Overview

This PR fixes #140 

Changes:

* Added some tests that ascertain that total pixel counts are consistent with categorized counts
* Chased down a bug that was globally summing pixel counts for multi-feature polygons
* Cleaned up the `crs_sniffer` to provide better functionality and logging

## Related Issue / Discussion
The `lu` dictionary was outside of the loop that examines and groups the categories for each feature. For multi-feature shapefiles, this was adding all the 

## Additional Information
There seems to be some weird behaviour with the CRS reprojection utility. I added a few more specific calls to vector and raster utilities as PROJ4 strings weren't being produced but GDAL/OGR was successfully guessing which projection to use. Logging is a bit clearer now but this needs to be looked into more in another issue.

*One thing worth mentioning is that having an OrderedDict under the entry "land-use" in the properties creates a non-standard GeoJSON, that can't be used for categorization purposes (ie: Cannot be parsed with ArcGIS, QGIS, (Geo)Pandas, etc.)* **Is there a reason why this data needs to be nested within the properties?**


tests/test_wps_generic_zonal_stats.py
```diff
@@ -100,3 +111,42 @@ def test_geoserver_dem_wcs(self):
...
+            service='WPS', request='Execute', version='1.0.0', identifier='zonal-stats', datainputs=datainputs)
+
+        assert_response_success(resp)
+        out = get_output(resp.xml)
+        feature = json.loads(out['statistics'])['features'][0]
+
+        stats = feature['properties']
+        assert {'count', 'min', 'max', 'mean', 'median', 'sum', 'nodata'}.issubset(stats)
+
+        # Check for accurate pixel counts
```
huard: This piece of code seems to be a copy-paste from above. I suggest making it into a function.

huard:
The aggregated land use could be a second independent output from the individual stats.

Zeitsperre:
> Let's convert `lu` to a real dict before saving to json.

This isn't so much an issue, as converting any type of dictionary (dict, DefaultDict, OrderedDict, etc.) results in the same data being written out; The issue is having a nested dictionary under a property entry.

Zeitsperre:
> The aggregated land use could be a second independent output from the individual stats.

Yes. I'm all for this solution.

huard:
Got it. Go for it.

tests/test_wps_generic_zonal_stats.py
Zeitsperre: Sounds good. I can make it a testing function under `common`.

huard:
There is at least one notebook that you'll need to change following the `lu` change.

raven/processes/wps_nalcms_zonal_stats.py
```diff
@@ -152,30 +159,36 @@ def _handler(self, request, response):
...
+            land_use = list()
             for stat in stats:
+                lu = defaultdict(lambda: 0)
                 prop = stat['properties']
 
                 # Rename land-use categories
                 for k, v in categories.items():
                     lu[v] += prop.pop(k, 0)
+                for k in summary_stats:
+                    lu[k] = prop.pop(k, 0)
```
huard: I don't like the idea of mixing the land-use count with other stats. To me these are two different objects. Also, then you need to jump through hoops to find the total number of pixels, because you need to exclude those stats. 
You want to be able to do `sum(lu.values())` and get the total number of pixels.

huard: Also, I thought we were going to aggregate the land use from all features in the output. Here you are passing a list.

Zeitsperre: That's fair and an easy change. We're saying that we don't want to save the nodata and/or nan values in the landuse data either?

huard: You've got a good point there. I suppose nodata and nan's have the same meaning as the other land-use entries, they count a fraction of the pixels. `count` would be out however. Does that make sense ?

Zeitsperre: From what i can tell, `nodata` and `nan` pixels do not count toward the `count` total. They are currently saved in the GeoJSON but they have no bearing on the most recent `land-use` outputs.

huard: Then let's keep them out of the lu.

Zeitsperre:
Good to merge?

huard:
I'm not hot for bundling "count" and the land-use categories in one single dict.

Zeitsperre:
Should I `pop` count into completely separate object and return that? Nested dictionaries aren't compatible for GeoJSON properties.

huard:
Wait, I think I may have reviewed a previous version of the code, my bad. Sounds good. Made a few tweaks to avoid modifying the output from zonal stats, so the raw outputs are still available to the users.

PR #175, opened 2019-10-21
Fix 174

### LINKED ISSUE #174: Create a notebook based GUI for region selection
## Description

A notebook GUI where users can display polygons from a WFS layer, click on a region and get the associated polygon.

Zeitsperre:
## Overview

This PR fixes #174 

Changes:

* Added a notebook that uses ipyleaflet and the RAVEN hydrobasins selection WPS
* Better handling for lonlat tuples sent via WPS

## Additional Information
Thanks to @davidcaron for his examples on how to set up ipyleaflet!

### DIFF
--- a/docs/source/notebooks/Region_selection.ipynb
+++ b/docs/source/notebooks/Region_selection.ipynb
@@ -1,48 +1,27 @@
 {
  "cells": [
   {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Region Selection and Map Preview with Ipyleaflet"
+   ]
+  },
+  {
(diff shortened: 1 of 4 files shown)


huard:
Nice ! 
Suggestions: 
- Add a text intro to the notebook: This is what we're going to do here. 
- remove print(version) in first cell unless there is a reason to do so. 
- Add text cell instructing users to move marker in Lac St-Jean before going further
- the dash-array contour is confusing, maybe use a smooth contour
- Remove plotting of additional watersheds
- Display the total area on the map ?

Zeitsperre:
Done.

PR #178, opened 2019-11-07
conda env: add psycopg2 to be able to use Postgres to log pywps requests in Prod

tlvu:
## Overview

conda env: add psycopg2 to be able to use Postgres to log pywps requests in Prod

When the following config is used when running the docker image:

[logging]
level = INFO
database=postgresql://user:pass@postgres/raven

"psycopg2" technically is only needed for production deployment so
potentially be added via individual `conda install` inside `Dockerfile`
instead.  I choose to add it to `environment.yml` so that during
dev/testing, we can also replicate "production environment".

It's always a best practice that dev/test environment mirror as much as
possible the production environment to smoke out all the relevant bugs.

## Additional Information

PR https://github.com/Ouranosinc/PAVICS/pull/173 had to disable Postgres for Raven because of this missing library.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -34,5 +34,6 @@ dependencies:
 - pandoc
 - ipyleaflet
 - pysheds
+- psycopg2
 - pip:
   - spotpy
--- a/requirements.txt
+++ b/requirements.txt
@@ -33,5 +33,6 @@ unidecode
 geojson
 pandoc
 ipython # needed for the parsing of Jupyter Notebooks
+psycopg2 # to use postgres to log pywps requests like in Prod
 # xclim It seems to be installed by pip in src/ and confuse pytest.
 


tlvu:
> Doesn't seem to be causing any fires.

Agreed, we've been running without Postgres in Prod since the beginning.

But according to @davidcaron it's much better to use a real DB in Prod.

Zeitsperre:
I would tend towards his judgment on this as well. Too many good reasons not to leverage an actual DBMS.

davidcaron:
I made tests a while ago, sending multiple fast requests using sqlite and postgres and I had issues with sqlite. It's supposed to be pretty good at handling concurrency, but I still had problems that I don't exactly remember unfortunately. It had to do with how you deployed the pywps wsgi app, whether using threads and/or processes.

PR #180, opened 2019-11-13
Fix notebooks for jenkins automated testing

tlvu:
## Overview

This PR fixes the notebooks so more are passing on Jenkins (still not all are passing).

Jenkins run: http://jenkins.ouranos.ca/job/PAVICS-e2e-workflow-tests/job/enable-raven-notebooks-in-jenkins/14/console

I think, not fully sure, that the only 3 root cause failure are:

* `progress=True`, need to somehow wait for it to finish or not use it

* ipyleaflet notebooks needs a default selection because there are no interactive user to make the selection (this is the one I am not sure)

* `Raven_run_parallel_basins.ipynb` failing with some broken pipe

Edit:

* add 3rd reason

### DIFF
--- a/docs/source/notebooks/Region_selection.ipynb
+++ b/docs/source/notebooks/Region_selection.ipynb
@@ -26,6 +26,13 @@
    ]
   },
   {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "If your `notebook` is version prior to `5.3`, you might need to run this command `jupyter nbextension enable --py --sys-prefix ipyleaflet`.  For more information see https://ipyleaflet.readthedocs.io/en/latest/installation.html."
+   ]
+  },
+  {
    "cell_type": "code",
    "execution_count": 3,
    "metadata": {},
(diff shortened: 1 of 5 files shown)


docs/source/notebooks/Region_selection.ipynb
```diff
@@ -25,26 +25,6 @@
...
-     "text": [
-      "Enabling notebook extension jupyter-leaflet/extension...\r\n",
-      "      - Validating: \u001b[32mOK\u001b[0m\r\n"
-     ]
-    }
-   ],
-   "source": [
-    "# Ensure that the ipyleaflet extension is enbaled\n",
-    "\n",
-    "!jupyter nbextension enable --py --sys-prefix ipyleaflet"
```
huard: This instruction should probably be moved somewhere, though.

tlvu: Humm, I could leave them as regular reminder kind of text but not executable code (it's breaking inside Jenkins run).

tlvu: Fixed in 3fe735b

PR #186, opened 2019-12-20
Add support for time shift and linear transform in rv templates

huard:
## Overview

First steps to support ERA5 hourly data.

Changes:

* Add time_shift and variable_linear_transform tags (necessary for hourly data)
* Add SuppressOutput tag to emulators
* By default, Ostrich will suppress output -> much faster
* Create symbolic links between identical rv files to reduce duplication
* Tag latitude, longitude, elevation in templates
*  Fix bug where parallel model instantiations with different parameters would affect each other
* Check netCDF units to make sure they are what Raven expects. 

## Additional Information


julemai:
Looks good to me. I very much appreciate that you got this implemented on such a short notice. Thanks!!

huard:
It needs to be merged before closing. Let's wait for the checks to complete though. 

Also, the changes have not been tested with hourly inputs yet, so there is a bit of magical thinking.

julemai:
I see. Sounds good!

huard:
Test pass, only need to update the notebooks.

huard:
@richardarsenault Could you take a look ? 
Main change is that Ostrich by default now does not return the hydrograph. You can get it back by setting `suppress_output=False`.

richardarsenault:
Everything looks good to me, code seems to do the right thing and tests are passing. However I see that the tests are running with daily time step files. I understand that this is the first step but do we have a Raven-ready hourly data nc file we could test with?

huard:
Good point. @julemai ?

julemai:
Yep, good point. I think hourly time shifts are only applied to daily data when they are multiples of 24. I don't have hourly data for the test catchment though.

huard:
We can probably test with opendap links. 
http://convection.meas.ncsu.edu:8080/thredds/dodsC/safaris/era5/PRECIP/12312009_18Z_18hr_precip.nc

richardarsenault:
@julemai is it easy for you to setup a test raven config using HMETS with random ERA5-hourly data? I could try and see if I can get the same on the server version using the same inputs.

julemai:
@richardarsenault I can try when I am back in office on Thursday. The guys who need the notebook said that they can also wait 2-3 weeks.

richardarsenault:
@julemai OK thanks. I've just hit a bit of a snag. RAVEN is telling me that the variable "tasmin" is inexistant in my merged t2m/pr netcdf file. Does RAVEN absolutely require tasmax and tasmin? Because by definition hourly data usually only has instantaneous data from which daily max and min can be dervied. How is that done in RAVEN? This will help me adjust my files consequently. Thanks!

julemai:
It sounds like you define your varname as “Tasmin” and “tasmax” in the rvt file and it seems like your variables are called differently. You need to adjust the varname to however they are called in your file. Raven will take care of up- and downscaling your variables to the temporal resolution needed. You could email me an rvt file and the NetCDF header and I make the adjustments if you would like to.

richardarsenault:
Ah, it's clear now, thanks! Ill try and figure it out myself, if I get stuck I'll reach out to you.

richardarsenault:
@huard I think the .rvt templates seem to require the {tasmax} and {tasmin} tags by default? Is there a way we can make this dynamic and/or add t2m while leaving the others blank?

julemai:
To clarify from the RAVEN perspective. The `rvt` either needs:
```
   :Data TEMP_MIN deg_C
      :ReadFromNetCDF
         :FileNameNC     {tasmin}
         :VarNameNC      {tasmin_var}
         :DimNamesNC     {nc_dimensions}
         :StationIdx     {nc_index}
      :EndReadFromNetCDF
   :EndData
   :Data TEMP_MAX deg_C
      :ReadFromNetCDF
         :FileNameNC     {tasmax}
         :VarNameNC      {tasmax_var}
         :DimNamesNC     {nc_dimensions}
         :StationIdx     {nc_index}
      :EndReadFromNetCDF
   :EndData
```

or 

```
   :Data TEMP_AVE deg_C
      :ReadFromNetCDF
         :FileNameNC     {tave}
         :VarNameNC      {tave_var}
         :DimNamesNC     {nc_dimensions}
         :StationIdx     {nc_index}
      :EndReadFromNetCDF
   :EndData
```

huard:
@julemai ERA5 uses and offset and a scale_factor. I guess I need to tweak the LinearTranform to account for this + the unit change ?

huard:
But it looks as if Raven is not always applying the LinearTransform, why? 
```
  if ( strcmp(DimNamesNC_stations.c_str(),"None") ) {

    // no switch of dim_order required because aTmp2D already points to
    // correct location in aVec data storage (see step 9)

    for (int it=0; it<ntime; it++){                     // loop over time points read
      for (int ic=0; ic<1; ic++){                       // loop over stations
        aVal[it] = LinTrans_a * aTmp2D[ic][it] + LinTrans_b;
      }
    };

  }
  else {

    // no switch of dim_order required because aTmp2D already points to
    // correct location in aVec data storage (see step 9)

    for (int it=0; it<ntime; it++){                     // loop over time points read
      aVal[it]=aTmp1D[it];
    };
  }
```

julemai:
@huard You are absolutely right. I resolved the bug and provide the revised code now under `http://www.civil.uwaterloo.ca/jmai/raven/raven-rev240.zip`

huard:
I'm getting different results for HBVEC. Expected ?

julemai:
That's likely. We fixed a lot of bugs- especially with NetCDF. I don't know if you use `:TimeShift`. That would make a difference too.

huard:
Not for this test case, but OK. 
Should I just set the expected diagnostic value to the new result I get ? 
```
    def test_simple(self):
        ts = TESTDATA['raven-hbv-ec-nc-ts']
        model = HBVEC()
        params = (0.05984519, 4.072232, 2.001574, 0.03473693, 0.09985144, 0.506052, 3.438486, 38.32455, 0.4606565,
                  0.06303738, 2.277781, 4.873686, 0.5718813, 0.04505643, 0.877607, 18.94145, 2.036937, 0.4452843,
                  0.6771759, 1.141608, 1.024278)

        model(ts,
              start_date=dt.datetime(2000, 1, 1),
              end_date=dt.datetime(2002, 1, 1),
              area=4250.6,
              elevation=843.0,
              latitude=54.4848,
              longitude=-123.3659,
              params=params,
              )

        d = model.diagnostics
        np.testing.assert_almost_equal(d['DIAG_NASH_SUTCLIFFE'], -0.075407, 4)
```
now yields 
```
AssertionError: 
Arrays are not almost equal to 4 decimals
 ACTUAL: 0.0186633
 DESIRED: -0.075407
```

julemai:
Let me check with James. Could you please email me the 5 rv* files and the forcings and observations for the test case that fails? This would help finding the reason faster. Merci!

julemai:
James is my hero. :)

He just looked at the outputs and knew what it is.

> Quick guess. Try adding the following command:
 `:SuppressCompetitiveET`
 to the .rvi and run with v240 . I see a shift in evaporation occurring first, so this could be it. Somewhere since v183, competitive ET has been the default, prior to that multiple ET processes could double dip.

I did that and the results are exactly the same then. It is however preferred to not `:SuppressCompetitiveET`. Hence the change in the HBV is ok and well explained.

I'd just change the expected NSE value for HBV-EC.

richardarsenault:
Thanks @julemai. I think we should go with 2 and simply make sure the difference is documented. No reason to keep a bug if we can improve I think.

huard:
Tests pass except for the notebooks. Will merge.

PR #194, opened 2020-02-20
Update raven to version to 2.9.2 (rev240)

huard:
## Overview

Update Raven. Fixes issue with LinearTransform not being applied to 1D gauge data. 

## Related Issue / Discussion

HBV-EC returns different diagnostic results.


julemai:
LGTM.

We are about to release a new version of Raven. It will be very close to rev240.

Cheers,
Julie

PR #200, opened 2020-03-05
Era5 notebook

richardarsenault:
## Overview
This PR includes a few notebooks, with one particular notebook "WF_FromShapefileToStreamflow.ipynb" that allows users to select the dataset (ERA5/NRCAN) and hydrological model (GR4JCN/HMETS/MOHYSE/HBVEC) and perform a complete simulation starting from just a shapefile.

The code makes use of WPS services to extract properties from within the shapefile (area, centroid lat/long, elevation) required to run RAVEN. It then uses the boundaries to extract the gridded data source selected and takes the spatial average of the points falling within the contour. The hydrological model is parameterized and run, providing streamflow stats and graphs.

Current limitations are:
1- Works for North America but West of 0° will be problematic for ERA5 due to 0:360° longitude scheme. Right now is hardcoded as longitude-360.
2- HBVEC and GR4JCN return an error due to the rain_snow_separation that seems to not be coded.
3- No streamflow provided, therefore calibration is not an option. Parameters are fixed and no NSE diagnostics are returned.
4- Had to install python package "salem", Don't know how to force that to be done automatically for future users during the deployment.

Also other notebooks are present in the PR that I wish to continue improving but are not necessarily required in the master at this stage...


richardarsenault:
Salem seems to be failing to parse CRS of the shapefile now, even though it worked yesterday. @huard can you test and see if it works for you? Do you have an ETA on when we'll be able to do this type of operation on the server side (as we'll eventually replace the code in these workbooks to use the methods on the server)? Thanks!

huard:
Working on it.

PR #204, opened 2020-03-09
Pytest for notebooks in allowed_fail build

### LINKED ISSUE #203: In Travis-CI, split the notebooks validation from the tests.
## Description

I'm willing to merge changes where notebooks fail if all unit tests pass. 

We'll confirm notebooks all pass before a release, but for the moment it's time consuming to update all notebooks.

Zeitsperre:
## Overview

This PR fixes #203 

Changes:

* Added a Travis CI build to explicitly test the notebooks and is allowed to fail.

## Additional Information

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,21 @@ matrix:
       env:
           - CONDA_FN="Miniconda3-latest-Linux-x86_64.sh"
           - DOCS=true
+    -   os: linux
+        name: "Linux (Python3.7 + notebooks)"
+        python: "3.7"
+        dist: xenial
+        addons:
+            apt:
+                packages:
+                    - libnetcdf-dev
+        env:
+            - CONDA_FN="Miniconda3-latest-Linux-x86_64.sh"
+            - NOTEBOOKS=true
+  allow_failures:
+    - env:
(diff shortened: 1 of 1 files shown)


huard:
That was fast ! 
Looks like it's working. Thanks. 

Have you heard of a way to reuse the conda environment from one test environment to the other ? This would probably speed up the tests. 

Will merge for now, but if you see something, please open a ticket to discuss it.

Zeitsperre:
Unfortunately, I don't think that's a feature in Jenkins/Travis CI. I think that the only way to do this would be to run the tests in one of the build and ignore the failing flags if raised. In a makefile, these are commands run with a hyphen, e.g.:
```
.PHONY: clean
clean: srcclean env_clean raven_clean ostrich_clean
	@echo "Cleaning generated files ..."
	@-for i in $(TEMP_FILES); do test -e $$i && rm -v -rf $$i; done
```
The same is true for Travis CI builds. We can try modifying an existing build to perform these tests, but we can't know if they fail unless we look closely at the logs.

All the same, I'll give it a quick look and keep an eye open. I agree that it would be useful to shave machine time off of some builds.

PR #218, opened 2020-03-17
Add :DontWriteWatershedStorage when suppress_output is True

### LINKED ISSUE #209: Add :DontWriteWatershedStorage parameter
## Description

Assign by default to calibration runs.

huard:
## Overview

This PR fixes #209 

Changes:

* Modified tag
* Modified tests
* Do not write empty zip files

Not obvious in the test suite that there is a speed up.


richardarsenault:
Works perfectly, and tag is considered, but indeed the speedup is not as much as expected. We'll have to look into that but this is not lost. Merging.

Issue #145, opened 2019-06-12
HMETS OSTRICH calibration does not return same NSE as an identical run using the optimized parameters.


richardarsenault:
Test can be performed in notebook: Running_OSTRICH_hmets

richardarsenault:
@julemai @huard : This can also be spotted in the test_emulators.py test, line 359 in HMETS_OST. It has not been resolved.

julemai:
I can have a look tomorrow. Which branch are you working in?

richardarsenault:
You can look at branch fix-222, test_emulators.py in HMETS_OST. Thanks!

julemai:
Just a quick update. I am looking into this since quite a while now. My best guess is that the derived parameters are not correctly handled.

I (again) don't know how to access the actual run folders on my Mac when a test is failing. Any idea how I do that. I run `make test` in my terminal.

julemai:
One thing I already found is that in `emulators.py` the following need to be adjusted:
```
        self.rvd['SUM_MELT_FACTOR'] = self.rvp.params.MAX_MELT_FACTOR   
        self.rvd['SUM_SNOW_SWI'] = self.rvp.params.SNOW_SWI_MAX     
```
instead of using:
```
        self.rvd['SUM_MELT_FACTOR'] = self.rvp.params.MIN_MELT_FACTOR + self.rvp.params.MAX_MELT_FACTOR 
        self.rvd['SUM_SNOW_SWI'] = self.rvp.params.SNOW_SWI_MIN + self.rvp.params.SNOW_SWI_MAX 
```   
because some derived parameters are already derived in `model.optimized_parameters` while others aren't.

julemai:
The `calibrated_params` of MOHYSE does not contain the 2 calibrated HRU values... So, the test for MOHYSE fails.

julemai:
All tests except one are now working. :) 

@huard : The error I am getting for `tests/test_emulators.py::TestMOHYSE_OST::test_simple` is the following:
```
Traceback (most recent call last):
  File "/Users/j6mai/Documents/GitHub/raven/tests/test_emulators.py", line 462, in test_simple
    params=model.calibrated_params
  File "/Users/j6mai/Documents/GitHub/raven/raven/models/base.py", line 417, in __call__
    procs = self.run(ts, overwrite, **kwds)
  File "/Users/j6mai/Documents/GitHub/raven/raven/models/base.py", line 410, in run
[...]

PR #258, opened 2020-04-28
Update Raven version [WIP]

### LINKED ISSUE #222: Update Raven to rev 251
## Description

- New version of Raven allowing for YYYY-MM-DD time units, assuming 00:00:00. 
- Bug fix to PET_OUDIN, expect changes to outputs

huard:
## Overview

This PR fixes #222, #145

Changes:

* Raven version is now 254

## Related Issue / Discussion
To rebuild: 
```
make raven_clean
make raven_dev
```

## Additional Information
Lots of failing tests... need to understand what changed. 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,6 +82,7 @@ src/
 testdata.json
 tests/*/*/processor_0
 tests/*/*/OstOutput*
+tests/*/*/*.nc
 
 # IPython
 .ipynb_checkpoints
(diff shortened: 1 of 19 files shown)


julemai:
The PET_OUDIN implementation changed since we found a bug in there. So, all models that use that PET option (in RVI) will give different results.

richardarsenault:
I've checked and fixed all failing tests except one: In test_emulators.py, there is a test for HMETS_OST where we calibrate HMETS (with fixed seed) and then rerun HMETS with the optimized parameter set. However there is a small difference in NSE, like if there was a difference of 1 day in the simulation, but the tests seem setup correctly. Need to investigate this further.

richardarsenault:
Fixed, now all passes.

huard:
I think we'd better merge this even though the travis build is broken.

richardarsenault:
I agree, merging.

PR #259, opened 2020-04-28
Climatology esp

### LINKED ISSUE #236: WPS for climatology-based ensemble streamflow predictions
TODO: Define API

richardarsenault:
*** Needs Raven revision 251+ to work ***

Fixes #236 

First implementation of the climatological ESP WPS. From observed climatological time-series, appropriate weather scenarios are derived and used as forecast scenarios for a specific user-defined forecast date. The WPS returns a netcdf file with N members and T time steps.

This version works and graphical analysis shows it does what it is supposed to. However we could eventually make it more efficient.

1- Right now, there is no model warm-up --> save hydrological states --> run forecasts from these initial states. Instead, the initial states are found by simply running the model from day 1 to the end of the forecast window by simply replacing the forecast weather at the end. So we might have to run a 10-year simulation and 30 day forecast for one member, then another time with the same 10-year simulation followed by the 30 days of the 2nd forecast, etc. So the result is the same but much less efficient that doing it straight away.

2- MOHYSE HRUs are not yet supported or tested.

3- general refactor of the kwds management is probably required, as is the netcdf variable management.

Again, the code works but will need to be improved. I suggest we keep that for maintenance.

### DIFF
--- a/raven/processes/__init__.py
+++ b/raven/processes/__init__.py
@@ -23,6 +23,7 @@ from .wps_q_stats import TSStatsProcess, FreqAnalysisProcess, FitProcess, BaseFl
 from .wps_indicator_analysis import GraphIndicatorAnalysis
 from .wps_graph_objective_function_fit import GraphObjectiveFunctionFitProcess
 from .wps_graph_fit import GraphFitProcess
+from .wps_climatology_esp import ClimatologyEspProcess
 
 processes = [
     RavenProcess(),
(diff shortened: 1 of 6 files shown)


richardarsenault:
I've thought about the forecasting workflows and we shouldn't implement this yet. We should first provide a generic forecasting method that takes a series of initial condition files (storage rv files?), parameters, forecast data time series and performs the forecast, no matter the source. I.e. this WPS would be the backbone of all forecasting methods (ESP-based, GeoMet, Caspar, other). Closing this PR.

PR #269, opened 2020-05-07
Implemented empirical forecast flood risk WPS

richardarsenault:
This WPS (and test suite) takes a forecast netcdf file and a flood level, and returns an empirical flood risk. The netcdf file can be either deterministic or ensemble:

Ensemble: The forecast netcdf file must contain a "member" coordinate. For each forecast day, we calculate the empirical probability of exceeding the flood level. Therefore if 4 members out of 10 exceed that flood level for one day, that day will have a value of 0.4 assigned to it. The returned netcdf file will contain one exceedence probability per day in the dataset.

Deterministic: If there is no "member" coordinate (or if there is one but it's size is 1), then the result is a similar netcdf file but whose responses are binary, i.e 1 if that forecast exceeds the flood level for that day and 0 otherwise. There is one value per day.


huard:
@richardarsenault 

The number of relevant lines of code / boilerplate ratio is not very high here. Could we bundle more content into this process (graphic, etc) ?

richardarsenault:
Yes absolutely, I think this is just the beginning but I would like to start with this and then add on to it. For starters, I am absolutely horrible at making graphics in python... also, eventually this will also tie into another type of analysis such as extracting fixed percentiles values of flow (i.e. what is the 5thm 25thm 50th, 75th, 90th, 95th and 98th percentile or something like that).

huard:
Ok, then I suggest you flesh out the API (inputs, outputs), leaving some of them blank for now and we'll implement them later.

huard (approved):

Issue #261, opened 2020-04-29
WPS to perform generic forecasts from forecast weather data and initial model states.


richardarsenault:
This WPS should take as inputs:

- initial hydrologic states (probably from the storage files but how to use them in hot-starting the model remains to be seen);
- forecast data as a netcdf with N-days x M-members for the required variables;
- model name and parameters
- other configuration tags such as linear transforms and time shifts, PET formulation, rain-snow split, latitude and elevation, etc required to run the raven model.

It will output the simulated forecasts as a N-days x M-members streamflow data netcdf. 

It should be generic in that it could then be used for the Climatology-ESP, Caspar and Geomet forecasts, as well as making assimilation and pre-processing more convenient.

richardarsenault:
@julemai do you have a good idea on how to use a previous run's storage files to hot-start another run?

julemai:
Raven is writing an output file `solutions.rvc`. This contains all the states of the last time step. Replace the current, for example, `raven-hmets.rvc` with the `solutions.rvc` and you will be able to start with the states you finished with.

richardarsenault:
@huard Could you look into this, how we can manage to do this without breaking the current setup of "when run Raven --> setup rv files" by substituting another one from a previous run?

huard:
1. Quick and dirty option: write a method that replaces the rvc file or template by the solution file.
2. Elaborate option: template initial values in emulators, parse the solution and feed the parameters to the template.

Option 2. will let also users specify initial conditions in general, which is currently not possible.

richardarsenault:
Option 2 will enable data assimilation, a huge part of future forecasting endeavors. So i would propose that if it's not TOO bad. Thanks Sent from my Samsung Galaxy smartphone.
-------- Original message --------From: David Huard <notifications@github.com> Date: 2020-04-29  16:01  (GMT-05:00) To: Ouranosinc/raven <raven@noreply.github.com> Cc: richardarsenault <richard.arsenault@etsmtl.ca>, Author <author@noreply.github.com> Subject: Re: [Ouranosinc/raven] WPS to perform generic forecasts from forecast weather data and initial model states. (#261) 

Quick and dirty option: write a method that replaces the rvc file or template by the solution file.
Elaborate option: template initial values in emulators, parse the solution and feed the parameters to the template.

Option 2. will let also users specify initial conditions in general, which is currently not possible.

—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or unsubscribe.
[...]

PR #270, opened 2020-05-08
Allow restarts and define initial conditions [WIP]

huard:
## Overview

This PR related to #261 

Changes:

* Added method `resume` to replace RVC with solution file.
* Started working on templating initial conditions [WIP]

## Additional Information


richardarsenault:
This is great! I like the model.resume() method. Question though: Can we pass a "model" instance through WPS? Like can I do a first pass, then call a WPS with the model state as an input to the WPS function to hot-start the model in the WPS?

huard:
No, but we could pass the solution file. This is not supported from WPS yet, but fairly easy (a ComplexInput `solution` to wpsio, then run resume on this file).

richardarsenault:
OK awesome. Can you setup this logic? I can pass the rvc file through wpsio etc., but can you pass the rvc file to the resume method?

huard:
Done, but untested.

richardarsenault:
I am having trouble finding an elegant way to test this without calling multiple functions. I have an Idea, maybe it's not too complex to add:

Could we simply add a parameter in the inputs to the hydrological model (such as ts, params, start_date, etc.) that would be "initial_state" and that would take a path to a file (and default to None)? That way all our current logic stays the same and we could simply pass that argument whenever we call the model. It would make it more streamlined I think, rather than setting up a model instance, THEN calling the resume method, etc.

Just an idea, if it's not possible or too complex I'll work with what we have! Thanks!

huard:
@richardarsenault See last commit.

richardarsenault:
IS it possible you forgot to push a file named "states.py" or something like that, because the line 6 in rv.py is failing:

from . state import HRUStateVariables, BasinStateVariables

No module named 'raven.models.state'

Does that sound plausible?

Thanks

huard:
Yes indeed. Will do.

Le dim. 17 mai 2020 12 h 05, richardarsenault <notifications@github.com> a
écrit :

huard:
@julemai Some rvc files have
```
:BasinInitialConditions
:Attributes, Q
:Units , m3/s
1 , 3.6
:EndBasinInitialConditions
```
But `Q` is not in the solution file per say (only `Qout`, `Qlat`, `Qin`).  

Is `BasinInitialConditions` part of `BasinStateVariables` or it's independent ?

julemai:
> @julemai Some rvc files have
> 
> ```
> :BasinInitialConditions
> :Attributes, Q
> :Units , m3/s
> 1 , 3.6
> :EndBasinInitialConditions
> ```
> 
> But `Q` is not in the solution file per say (only `Qout`, `Qlat`, `Qin`).
> 
> Is `BasinInitialConditions` part of `BasinStateVariables` or it's independent ?

I talked with James. He said "`Qout` is the same as `Q` in this context. " and "`:BasinInitialConditions` is like the low-power version of `:BasinStateVariables`. It only overwrites the instantaneous flow Qout/Q".

PR #287, opened 2020-07-15
Update version of xarray and xclim

### LINKED ISSUE #286: Xclim SDBA timeout with NRCan
## Description

Trying to implement bias-correction of climate data using Xclim to push bias-corrected data to a raven model, to do climate change impact assessments.

The frame seems to be working, able to bias correct climate model data based on ref, fut vs NRCan as the historical data.

However, there seems to be a problem with the size of the ncml and the timeout if I use more than 10-12 years as the historical data. I get a : "Netcdf: DAP failure" error if I use too many years.

Also, will the new version of xarray (master) be pushed on the Raven GIT so we can use xclim directly or would that cause too much trouble? Just so we don't have to always switch environment.

huard:
## Overview

This PR partially fixes #286 

Changes:

* Pin xarray==0.16
* Pin xclim==0.18

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -15,8 +15,8 @@ dependencies:
 - psutil
 - numpy
 - scipy
-- xarray==0.13
-- xclim==0.12.2
+- xarray==0.16
+- xclim==0.18
 - rioxarray
 - matplotlib
 - dask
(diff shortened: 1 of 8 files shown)


Zeitsperre:
There's still an error associated with the newer version of scipy. @huard or @richardarsenault, care to jump in?

huard:
Thanks @Zeitsperre for looking into this. I'll be able to get into it later today.

richardarsenault:
Will check tomorrow, have access problems to my linux server right now

PR #292, opened 2020-07-22
Bias correction

richardarsenault:
Adds interacton with Xclim for bias correction with a simple test suite.


richardarsenault:
@tvlu can you check for the conflicts please?

.travis.yml
```diff
@@ -56,6 +52,7 @@ jobs:
 branches:
   only:
     - master
+    - bias_correction
```
tlvu, line 55: Why do we want to limit Travis-CI to only `master` branch, then we have to add each dev branch manually, like this `bias_correction` branch? @Zeitsperre (b5a356609228a44a04928f91af5e5a76773546ef)

.travis.yml
```diff
@@ -75,8 +72,9 @@ install:
 
 before_script:
 #  # Start WPS service on port 5000 on 0.0.0.0
-#  - raven start --daemon --bind-host 0.0.0.0 --port 5000
-  - bash -c "source $HOME/miniconda3/bin/activate raven && make start"
+    #  - raven start --daemon --bind-host 0.0.0.0 --port 5000
+    - bash -c "source $HOME/miniconda3/bin/activate raven && pip install git+https://github.com/pydata/xarray.git@master"
```
tlvu, line 76: Is this install from xarray.git@master still necessary with xarray 0.16 from `master` branch? @Zeitsperre  (a125ff4bdd86881239226ac77dcf54ee721ad8b1)

Zeitsperre: With xarray at 0.16, the sdba functions that Pascal was working on are integrated, so no. This workaround can blissfully be thrown in the trash.

.travis.yml
Zeitsperre: This can be removed. It was placed in order to trigger the bias_correction branch that PRs were merging to in a few instances. The reason why I set only to master is that otherwise we tend to get two duplicate sets of builds; one that is the pushes set against master and another that is the pushes set against the last version of master it stemmed from.

Since we only care about the current state of master, as all pushes must integrate these changes, there's nothing to gain except more redundant build checks.

tlvu: I think the redundant build checks on `master` branch will happen only once, when the PR is merged to `master`.

Whereas instructions to build only the `master` branch will not test any commits on any devel branches and force all devel branches to hardcode the devel branch name like we just saw.

I think having the redundant build on `master` is a less bad of the 2 problems.  Wonder what's your opinion about removing this "build only `master` branch" instruction for all birds?

tlvu:
@richardarsenault I've deleted branch `bias_correction`.  It's a good habit to always delete the associated branch after the PR is merged to avoid leaving non useful branch behind.

You can create a new branch to continue working on your feature.  You can even re-create the same branch name if you wish.  At least it'll be starting at the good new location.

richardarsenault:
OK thanks, I am always kind of reticent of deleting branches and stuff so I usually do it a few weeks later to make sure we don't need anything still, but I guess that is futile, isn't it :) 

I will take note to delete after merging. Thanks for your patience with me :)

tlvu:
No worries.  Git basically uses branches to keep track of commits.  If a commit and its ancestors have no branch pointing to it, it'll be eligible to be garbage collected.  By merging to `master`, the `master` branch will be the one keeping those commits "alive".  That's why there is always at least one branch `master` in any git repository.

From the point of view of us the developers, once we merge to `master`, we can always navigate the history on `master` to find the previous branch if needed.  And when we want to continue a feature, we should always start from the head of `master` and not from the head of our previous branch because we risk falling behind `master`.  We always want to be up-to-date with `master` to integrate with new code from others, to ensure they do not break us and we do not break them.

Hope this further re-assure you that it's safe to delete a dev branch after it has been merged to `master` or even another branch, as long as there is at least one branch referencing those commits.

richardarsenault:
Sounds good, thanks!

PR #302, opened 2020-10-28
Add support for non-standard calendars. requires testing

### LINKED ISSUE #300: Detect input file calendar and write to config
## Description

If the netCDF input series uses a non-standard calendar, this information must be provided in the RVI config file. 

TODO: 
- [ ] Detect calendar from input files
- [ ] Ensure calendar consistency across inputs
- [ ] Write the calendar into the RVI config

huard:
## Overview

This PR fixes #300 

Changes:

* Added `calendar` property in RVI class
* Detect `calendar` from netcdf files
* Write calendar into RVI templates for all four emulators. 

** Needs to be tested with forcing using non-standard calendars before merging **

### DIFF
--- a/docs/source/notebooks/gridded_data_subset.ipynb
+++ b/docs/source/notebooks/gridded_data_subset.ipynb
@@ -6,7 +6,7 @@
    "source": [
     "# Running a hydrological model over a watershed defined by a shapefile\n",
     "\n",
-    "This notebook shows how to run Raven over a user-defined watershed. The watershed contour is provided by a shapefile, which we use to subset meteorological data and to extract watershed physiographic properties. The meteorological data is spatially averaged, then fed to the Raven hydrological model to simulate streamflow. "
+    "This notebook shows how to run Raven over a user-defined watershed. The watershed contour is provided by a shapefile, which we use to subset meteorological data and to extract watershed physiographic properties. The meteorological data is spatially averaged, then fed to the Raven hydrological model to simulate streamflow."
    ]
   },
   {
@@ -64,9 +64,9 @@
     "vec = TESTDATA['watershed_vector']\n",
     "print(\"The file location is: \" + str(vec))\n",
     "\n",
-    "# Choose a dataset to use. We have 'NRCAN' and 'ERA5' for now. \n",
+    "# Choose a dataset to use. We have 'NRCAN' and 'ERA5' for now.\n",
     "# NRCAN is only available in Canada, while ERA5 is global.\n",
-    "dataset = 'ERA5' \n",
+    "dataset = 'ERA5'\n",
     "\n",
     "# Choose a hydrological model to use. We have 'HMETS', 'GR4JCN','MOHYSE' and 'HBVEC'.\n",
     "hydromodel = 'HMETS'"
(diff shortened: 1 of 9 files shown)


richardarsenault:
Thanks for this. The calendar tag seems to be working, but I think something else needs to be done to fix the "duration" tag. As of now, it still indicates 3651 days, although there are clearly 3650 in 10 full 365-day calendar years. Possible that there is a calculation being done here that needs to be updated too? 

I've attached the rvi file of a recent run showing this problem (renamed as .txt to upload here):

[raven-hmets.txt](https://github.com/Ouranosinc/raven/files/5461396/raven-hmets.txt)

If you have any ideas I'm all ears!

julemai:
Maybe it's just easier to set the `:EndDate` instead of the `:Duration`?

richardarsenault:
Perhaps? All I can say is that I know we can call the model with either endDate OR duration, and in my case I used an endDate so there was probably a conversion from endDate to Duration somewhere in the code?

julemai:
ok. I see. You are talking about the PAVICS code and not Raven itself, right?

richardarsenault:
Yes, exactly. Sorry for the confusion!

julemai:
All fine. Just checking... :)

huard:
@richardarsenault Now ?

richardarsenault:
Yes sir, works as advertized! Thanks! Can we merge this now so I can more easily merge master into my branch?

huard:
I suggest we confirm all tests pass before merging. You can merge this branch into your own branch however. `git merge fix-300`.

huard:
The test for bias correction is failing... unclear why. Data access issue ?

richardarsenault:
Tests are now all passing except test_version, which I suspect is minor.

richardarsenault (approved):
Everything works OK, test_version is failing due to a... version change somewhere, but it seems to be a simple assert.

huard:
I'm not seeing this version error here. Did you install an updated version of raven ? Also, I'm getting errors related to wps_forecast_evaluation. Is it possible that you updated test files locally (XSS) and did not commit them ?

richardarsenault:
hmmm I'll check, it's possible as I did play with the forecasting tests, but I for sure didn't merge my xss branch in this one, so I'm unsure how it got to be different between our versions? I'll look into it today.

richardarsenault:
OK so I just checked and here is my status:
1- It seems I am running Raven 2.9.2 while version 3.0 is expected, so I'll update that later.
2- The xss branch was merged a while ago, and my version in fix-300 is exactly the same as in master, and I have not changed anything to that code in a while now. Could you give more details on what is failing so I can debug a bit more?

huard:
On this page click on "Details" for "Travis CI - Branch "just below "All checks have failed". Pick the Python3.6 test suite and scroll to see the test output.

richardarsenault:
OK, so the error is that: 

In wps_forecast_evaluation, line 125:
raise ValueError(f"{dim!r} not found in array dimensions {self.dims!r}")
ValueError: 'member' not found in array dimensions ('time',)

I've checked the ensemble forecast data netcdf (in testdata/XSS_forecast_data) and it's exactly the same between all branches.

Is there a way we can check on that server exactly what is happening? like add a breakpoint and check what the variable values are?

huard:
I get the same error locally, so I'll be able to test this out. What about the obs ?

richardarsenault:
The obs I did not test because the error comes from (seemingly) the fact that the ensemble forecast file does not have a 'member' dimension... but I just did a pull and test and everything looks good, and I have no "non-added" files in my repo.

richardarsenault:
Also, the deterministic forecast check works, and both the deterministic and ensemble use the same obs, so I am confident it is not the obs.

huard:
ok, so let's merge this PR and try to fix the issue later. I think I'm getting the same error with master.

richardarsenault:
OK sounds good

PR #306, opened 2020-11-07
Bias correct notebook

richardarsenault:
## Overview

This PR adds a bias-correction notebook that uses xclim to bias-correct climate model variables and feeds the corrected value into a hydrological model to produce a hydrograph.


huard:
I suggest to remove the first part with the vector handling and focus on the bias correction. I think it's easier to follow when tutorials are short and focused, rather than trying to do everything.

richardarsenault:
@huard Just to be clear - There are 2 notebooks here, one is the Full-process notebook and the other is the bias-correction notebook. Your comment concerns the bias-correction notebook, right?

So what I can do is remove the auto-extraction of area/latitude/etc. and just keep the shapefile geometry part, and provide numerical values as we do at other times, is that good?

huard:
Correct. 
Yes!

PR #315, opened 2020-11-10
Add test to exercise initial state updates

### LINKED ISSUE #314: Allow resuming at other dates than the timestamp date
## Description

The solution file includes a timestamp at the end of the simulation. However, for assimilation and initialization, it is necessary to be able to use the solution file to restart the run at an earlier time.

huard:
## Overview

This PR fixes #314

Changes:

* Added two tests, everything seems to work as expected.


richardarsenault (commented):
The new tests seem to work well and I will use the syntax you use in my code now, see if that helps. However, with this test suite I now get errors:

test_emulators.py::TestGR4JCN::test_parallel_params - AssertionError: assert 5 == 10

Don't know what's causing this new behaviour?

huard:
I reinitialized the list of config files because it contained duplicates when re-running the same model, but did not realize it would affect parallel computations. Removed the change.

richardarsenault (approved):
All good on my end!

PR #320, opened 2020-11-24
add deaccumulate option

### LINKED ISSUE #319: Add deaccumulate option
### RVT
The `:Deaccumulate` command, if included will take cumulative precipitation values stored in the NetCDF file (such as are output by some weather models) and convert these to precipitation rates, in mm/d.

huard:
## Overview

This PR fixes #319 

Changes:

* Added `:Deaccumulate` option to RVT template.

### DIFF
--- a/raven/models/base.py
+++ b/raven/models/base.py
@@ -542,8 +542,8 @@ class Raven:
                                                    var_name=alt_name,
                                                    dimensions=ds[alt_name].dims,
                                                    units=ds[alt_name].attrs.get("units"),
+                                                   deaccumulate=ds[alt_name].attrs.get("GRIB_stepType", "") == "accum",
                                                    )
-
                                 break
         return ncvars
 
--- a/raven/models/base.py
+++ b/raven/models/base.py
@@ -542,8 +542,9 @@ class Raven:
                                                    var_name=alt_name,
                                                    dimensions=ds[alt_name].dims,
                                                    units=ds[alt_name].attrs.get("units"),
-                                                   deaccumulate=ds[alt_name].attrs.get("GRIB_stepType", "") == "accum",
                                                    )
+                                if "GRIB_stepType" in ds[alt_name].attrs:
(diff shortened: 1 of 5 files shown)


raven/models/rv.py
```diff
@@ -309,6 +311,15 @@ def linear_transform(self, value):
 
         self._linear_transform = value
 
+    @property
+    def deaccumulate(self):
+        """Convert cumulative precipitation to precipitation rate (mm/d)."""
+        return ":Deaccumulate" if self._deaccumulate is True else ""
```
cjauvin, line 317: Vraiment un nitpick pour nitpicker ici mais le `is True` n'est pas redondant?

cjauvin (approved):

raven/models/rv.py
huard: Je voulais m'assurer que c'est bien un boolean, et pas juste 42.

cjauvin (approved):

raven/models/base.py
```diff
@@ -542,8 +542,8 @@ def _assign_files(self, fns):
                                                    var_name=alt_name,
                                                    dimensions=ds[alt_name].dims,
                                                    units=ds[alt_name].attrs.get("units"),
+                                                   deaccumulate=ds[alt_name].attrs.get("GRIB_stepType", "") == "accum",
```
richardarsenault, line 545: This line makes it that I cannot force the "deaccumulate" flag in model.rvt.pr.deaccumulate=True, if the flag is not in the file it returns "" and overwrites my <True> tag. Up until here, the flag is set to true, but then gets overwritten here. So I guess the best alternative is to add that flag when I generate the new netcdf for Raven from the forecasts?

richardarsenault: Actually, I did just that, adding the GRIB_stepType attribute to 'accum' got it detected and works now.

huard: Hum, but ideally we would not overwrite it. I'll patch something.

richardarsenault:
This now works. I will merge now so I can use it in my hindcasting branch.

Issue #185, opened 2019-12-06
Sync testdata folder to THREDDS testdata/raven


huard:
So we can run tests entirely on the platform.

huard:
Related: https://github.com/Ouranosinc/xclim/issues/525
We could use the same strategy of creating a stand-alone repo for test data. Might make the sync process cleaner.

tlvu:
I am trying to wrap my head around what's the real root problem here.

Basically we do not want tests to use relative path to the `testdata` folder in this repo.  How about just turn that relative path to an http path directly to github (ex: https://github.com/Ouranosinc/raven/raw/master/tests/testdata/hydro_simulations/raven-gr4j-cemaneige-sim_gr4jcn-0_Hydrographs.nc)?  This way, no other server is needed **and** it makes the `.ipynb` file standalone?

For optimization, we can detect if the relative path exist (dev full checkout mode), use that one and keep the http path as fallback only (tutorial mode).

I rather our Thredds server do not become a single point of failure for running tests.

huard:
1. I think we want to split the test data from the code. This would mean creating a `raven-testdata` repo. 
2. Some of the tests and tutorials should exercise DAP URLs, so for some tests, we'll need a THREDDS server. 

One idea is to have a github `raven-testdata` repo that gets synced on THREDDS when there is a release of `raven-testdata`. Then, we'll need a client API to fetch data either from THREDDS or from github.

tlvu:
>  2\. Some of the tests and tutorials should exercise DAP URLs, so for some tests, we'll need a THREDDS server.

Maybe Raven spawn its own Thredds server, like Emu does https://github.com/bird-house/emu/blob/56def2684fc28fee09089382de192075f065f3f2/docker-compose.yml#L12-L23 ?

Again I don't want all the tests to suddenly fail on Travis-CI and all local dev workstation just because Thredds is down for maintenance.  And let's say there is a new or updated dataset that is not yet synced to Thredds, how can a dev continue his work?

So my point is yes we'll still need to sync all the data to Thredds for tutorials but day-to-day dev workflow should not rely on Thredds.

Which test(s) need a DAP link right now?  Some data is already manually on Thredds?

tlvu:
Can Intake provide what we need "abstract the backend storage (local file, http link, dap link)"?

huard:
I'm wary to make the development environment more complex than it is, but I think you raise valid issues. 
[...]

PR #322, opened 2020-11-26
First pass at a get_file() function

Zeitsperre:
## Overview

Changes:

* Added a function for collecting data files and caching them or, in the case of OPeNDAP links, opens and provides the data directly.

## Related Issue / Discussion

https://github.com/Ouranosinc/raven/issues/185


raven/tutorial.py
```diff
@@ -0,0 +1,118 @@
...
+
+    if suffix == "nc":
+        try:
+            ds = _open_dataset(local_file, **kwds)
+            if not cache:
+                ds = ds.load()
+                local_file.unlink()
+            return ds
+        except OSError:
+            raise
```
huard, line 113: I suggest `get_file` is only responsible for getting the file from upstream or the cache. Then could have another open_dataset() function that calls it and returns the corresponding `Dataset` object.

Zeitsperre:
@huard this is good to be merged

huard (approved):
Nice !
How do you see the next step ? 
Is it to replace the links in `common.TESTDATA` by the urls ? Or use the links directly in the test files, without going through TESTDATA, or create pytest fixtures ?

Zeitsperre:
> How do you see the next step ?
> Is it to replace the links in `common.TESTDATA` by the urls ? Or use the links directly in the test files, without going through TESTDATA, or create pytest fixtures ?

The `get_file` approach helps us not have to deal with URLs and adding them in would likely be more complicated (as access via URLs is managed already, and examining the cache should be the priority). I would probably just replace from `from xarray import open_dataset` to `from raven.tutorial  import open_dataset` in our tests. We don't need to do much.

Issue #313, opened 2020-11-10
Sync ensemble forecasts on THREDDS


huard:
## Description

ECCC produces ensemble forecasts and distributes them through GeoMET and their datamart. The original plan was to fetch data from GeoMET using WCS directly in netCDF format, but the metadata fields are empty and there is no scheduled fix for this. 

The solution for now would be to fetch files from the "datamart". 

Create cron job to do the following every day at 7h and 19h.

- Copy to temporary directory files from 
   - https://dd.weather.gc.ca/ensemble/geps/grib2/raw/??/???/CMC_geps-raw_TMP_TGL_2m_latlon0p5x0p5_*_P*_allmbrs.grib2 
   - https://dd.weather.gc.ca/ensemble/geps/grib2/raw/??/???/CMC_geps-raw_APCP_SFC_0_latlon0p5x0p5_*_P*_allmbrs.grib2
- Convert files to netCDF(see below)
- Copy netCDF files on THREDDS to `Birdhouse/eccc/ensemble/geps`
- Clean old netCDF files

The `??/????` in the path refer to the 0-h of the forecast (00 or 12), and the forecast lead time (increments of 3-h then 6-h). 

Also (Travis)
- Create NcML to aggregate variables and time. Note that there are two *time* dimensions, one for the time the forecast was issued, and another for the forecast time. 

If this all works fine, we would follow a similar model to host other ensemble products (REPS, GDPS, RDPS).

### Notes
`CMC_geps-raw_TMP_TGL_2m_latlon0p5x0p5_2020110900_P000_allmbrs.grib2`
`_____________VAR______________________YYYYMMDDHH_PFTS_________`

where `YYYYMMDD` is the date and time the forecast was made, and `FTS` is the forecast lead time. Now that for precipitation, the first forecast is at 003 because it is the cumulative precipitation from time 00. 

Note that there are 65 3-h time steps followed by 25 6-h time steps. 

### Conversion to netCDF
$ pip install xarray cfgrib

```python
import xarray as xr

def convert(fn):
  """Convert grib2 file to netCDF format.

  TODO: Add parameter to change output directory.
  """
  ds = xr.open_dataset(fn, engine="cfgrib", backend_kwargs={'filter_by_keys':{'dataType': 'cf'}})
  ds.to_netcdf(fn.replace(".grib2", ".nc"))
```

tlogan2000:
@huard python code for retrieving and converting should go in `utilities`?

huard:
[...]

PR #321, opened 2020-11-26
Hindcasting from CASPAR data

richardarsenault:
Hindcasting is now possible from Caspar.

Uses the Caspar GEPS data on THREDDS to run a 20-member forecast on 10 days lead-time.

For now, the code uses a bogus initial state run for development, eventually we can use actual observed data to ensure good initial states. We can also, later, add REPS, GDPS and RDPS to have the complete set.

The code uses the deaccumulation option of fix-319 by forcing the attribute in the netcdf generated for the subset forecast data.

PR #323, opened 2020-11-27
Wps geomet

richardarsenault:
This PR is a first implementation of the ECCC datamart data aggregation forecasts from #313 .

Forecasts are currently limited to GEPS, should be no issue in extending to other models later.

Should work as long as the data keeps being updated through the ncml files on THREDDS. Raises an exception if there is no data to be found in the past 7 days.

branch name says 'wps_geomet' but that was a different time... everything related to geomet has been removed, and there is currently no WPS, only direct calls are implemented through the test function.


raven/utilities/forecasting.py
```diff
@@ -173,3 +182,98 @@ def perform_climatology_esp(model_name, forecast_date, forecast_duration, **kwds
...
+
+        # Find the most recent data on the server by looping from today and go
+        #backwards 5 days. If no data is present, not much value in forecasting.
+        filein=None
+        for i in range(0,7):
+            
+            # Write the string
+            date=date_now-dt.timedelta(days=i)
+            datefcst=dt.datetime.strftime(date,"%Y%m%d")
+
```
huard, line 212: I suggest we replace this by a `latest` NcML link as the default.

raven/utilities/forecasting.py
```diff
@@ -173,3 +182,98 @@ def perform_climatology_esp(model_name, forecast_date, forecast_duration, **kwds
     qsims["member"] = (["member"], avail_years)
 
     return qsims
+
+
+def get_recent_ECCC_forecast(region, climate_model='GEPS'):
```
huard, line 187: This function should be split in two: 
1. Get netCDF resource from PAVICS THREDDS, run health checks. 
2. Compute region average. 

Step 2 could be refactored in a couple of months to use  pangeo-netCDF. This would require CF coordinate bounds are included in the variables. An advantage is that averaging weights can be stored and reloaded to speed things up.

raven/utilities/forecasting.py
```diff
@@ -173,3 +182,98 @@ def perform_climatology_esp(model_name, forecast_date, forecast_duration, **kwds
...
+            times.append(single_time)
+
+    elif climate_model is 'GDPS':
+        filein='https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/dodsC/birdhouse/caspar/daily/GDPS_' + dt.datetime.strftime(date,'%Y%m%d') + '.nc'
+
+    elif climate_model is 'RDPS':
+        filein='https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/dodsC/birdhouse/caspar/daily/REPS_' + dt.datetime.strftime(date,'%Y%m%d') + '.nc'
+
+    elif climate_model is 'REPS':
+        filein='https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/dodsC/birdhouse/caspar/daily/RDPS_' + dt.datetime.strftime(date,'%Y%m%d') + '.nc'
```
huard, line 244: Use string template instead of `if`.

raven/utilities/forecasting.py
```diff
@@ -173,3 +182,98 @@ def perform_climatology_esp(model_name, forecast_date, forecast_duration, **kwds
...
+    shp = list(tmp.glob("*.shp"))[0]
+    vector = fiona.open(shp, "r")
+    shdf = [vector.next()["geometry"]]
+
+    # extract the bounding box to subset the entire forecast grid to something
+    # more manageable
+    lon_min=vector.bounds[0]
+    lon_max=vector.bounds[2]
+    lat_min=vector.bounds[1]
+    lat_max=vector.bounds[3]
```
huard, line 258: Convert to utility function `get_shp_bbox`.

raven/utilities/forecasting.py
huard: Step 2 would be implemented into clisops, on which xclim relies for subsetting and regridding operations.

huard:
@richardarsenault If you agree with the suggestions, @cjauvin will take care of making the modifications.

richardarsenault:
OK yes, totally makes sense. Not sure about the "latest" ncml. Is that like an Ncml that points to the most recent forecast available? If that is possible to get then yes, I think that would be the best!

For the subfunctions, also totally agree.

Not certain about the "use string template"? If its clear to @cjauvin then by all means, go ahead!

Also no idea what clisops so if that's something you can simplify, please go ahead!

raven/utilities/forecasting.py
```diff
@@ -18,6 +18,15 @@
...
 
+import rioxarray
+import fiona
+
+import tempfile
+from pathlib import Path
+from zipfile import ZipFile
+from xclim import subset
+
+import pdb
```
Zeitsperre, line 29: This should eventually be removed when the code is more stable

raven/utilities/forecasting.py
```diff
@@ -173,3 +182,98 @@ def perform_climatology_esp(model_name, forecast_date, forecast_duration, **kwds
...
+            # try to open the files. If it works, great, if no
+            try:
+                filein='https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/dodsC/birdhouse/testdata/geps_forecast/CMC_geps-raw_latlon0p5x0p5_' + datefcst + '00_allP_allmbrs.nc'
+                ds=xr.open_dataset(filein)
+                # Checking that these exist. IF the files are still processing, possible that one or both are not available!
+                assert 'pr' in ds
+                assert 'tas' in ds
+                break
+            except:
+                filein=None
```
Zeitsperre, line 222:
```suggestion
```suggestion
                if hasattr(ds, "pr") and hasattr(ds, "tas"):
                    break
                raise AttributeError
            except AttributeError:
                filein=None
```

Zeitsperre (commented):
Just a few comments. You might want to make use of the Pathlib library to simplify your `filein` handling.

raven/utilities/forecasting.py
cjauvin: @tlogan2000 It is my understanding that you will be in charge of creating this view so please ping me when it's ready so that I can complete the required changes for this PR.

richardarsenault: @cjauvin here you can see my limited programming proficiency, as announced... sorry for the extra work and thanks for the support!

tlogan2000: > @tlogan2000 It is my understanding that you will be in charge of creating this view so please ping me when it's ready so that I can complete the required changes for this PR.

sounds good

cjauvin: @richardarsenault It's a pleasure to help, and I'm currently thinking that I really have much to learn, in hydrology and probably a bunch of other stuff!

raven/utilities/forecasting.py
cjauvin: @huard @richardarsenault I just realized that the control flow is a bit weird here:

https://github.com/Ouranosinc/raven/blob/wps_geomet/raven/utilities/forecasting.py#L237-L244

as `filein` is not being used further, with models `GDPS`, `RDPS` or `REPS` (the last 3 `elif` clauses).

Shouldn't the creation of `ds`:

https://github.com/Ouranosinc/raven/blob/wps_geomet/raven/utilities/forecasting.py#L216

be pushed farther, i.e. at least after line 245:

https://github.com/Ouranosinc/raven/blob/wps_geomet/raven/utilities/forecasting.py#L245

raven/utilities/forecasting.py
```diff
@@ -173,3 +182,98 @@ def perform_climatology_esp(model_name, forecast_date, forecast_duration, **kwds
...
+
+    # extract the bounding box to subset the entire forecast grid to something
+    # more manageable
+    lon_min=vector.bounds[0]
+    lon_max=vector.bounds[2]
+    lat_min=vector.bounds[1]
+    lat_max=vector.bounds[3]
+    
+    # Subset the data to the desired location (bounding box) and times
+    ds=subset.subset_bbox(ds, lon_bnds=[lon_min, lon_max], lat_bnds=[lat_min, lat_max]).sel(time=times)
```
cjauvin, line 261: Another thing with the control flow I'm not sure I understand: if `climate_model` is not `"GEPS"`, `times` is not defined, so this will crash.

raven/utilities/forecasting.py
cjauvin: Question: are `pr` and `tas` supposed to be present only in the `GEPS` dataset, or in all? (GDPS, RDPS, REPS)

richardarsenault: they will be in all the climate model files.

raven/utilities/forecasting.py
richardarsenault: yes, it's because we have no files for these models yet. Times will be added when we get that info. For now, only GEPS will work though.

cjauvin:
ok in that case I guess that the flows involving the unavailable data
should be simply commented out, for better clarity and safety?

raven/utilities/forecasting.py
richardarsenault: Perhaps, yes! This was done at first because I had to cycle through the previous files to find an existing one. Now, we have the "latest" file which should ne used instead, so the opening of fileIn and checking of the presence of "tas" and "pr" in the file could go after this block.

richardarsenault:
@cjauvin yes probably a good idea at this stage!

raven/utilities/forecasting.py
richardarsenault: also, I should mention that the only 4 options that will ever be available are GEPS, REPS, GDPS and RDPS. so perhaps we should have a sanitizer or checker of some sort to make sure that this is the case. (but only GEPS for now).

raven/utilities/forecasting.py
cjauvin: @huard @richardarsenault I have completed my refactoring, but the exact link that needs to be used for the latest view is still in limbo. There's also the question of what dataset we want to use for testing, as using something transient for testing is not ideal, but OTOH having a 2.5 GB file in the testing data is not great either.

raven/utilities/forecasting.py
cjauvin: I have refactored into two functions.

raven/utilities/forecasting.py
cjauvin: At first I wanted to do it in `gis.py` but it seems to dangerously close to: https://github.com/Ouranosinc/raven/blob/master/raven/utilities/gis.py#L151

raven/utilities/forecasting.py
cjauvin: For my local testing I have downloaded a nc file (as you can see in the code, that needs to be changed) but it's not ideal because I then need to mess with the time series selection dates, etc.

richardarsenault: I think it should be here: https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/dodsC/birdhouse/testdata/geps_forecast/GEPS_latest.ncml

It seems to have been updated recently so I would guess this is the current location for the file? Unless you need something else. I can provide a much smaller file (by subsetting lat/lon) that could be added to testdata and that would be fixed in time, if that's what you mean?

cjauvin: This link is from testdata so I don't think it will be the final, official one.

About your smaller file proposition: yes I think it would be great if it's possible! One thing that we would need to be sure of though is the fact that whatever test we are doing with the smaller dataset offers the same guarantees of correctness than we would have running on the real, bigger dataset.

Also: I'm not sure how `times` should be computed:

https://github.com/Ouranosinc/raven/blob/wps_geomet/raven/utilities/forecasting.py#L244-L246

What should be the starting point?

richardarsenault: OK, I can make a smaller file, but it would be static (i.e. for a fixed date, so the tests would have to check that file in particular rather than a "live" version. But I think this would be OK if we pair it with a second test that checks to see if the real "latest" file is actually present, contains the correct variables and times. What do you think? The latter would not need to be included in the testdata DB.

Also, for the times: the thing is that the weather forecasts are not uniform. at first they are hourly, then 3-hours and towards the end they are 6-hourly. This is because a climate model is able to be relatively accurate on the hourly scale for 2-3 days, after that it loses in accuracy so going to 3-hours offers the same level of accuracy but with fewer data. Same reasoning for the jump from 3 to 6 hours. But RAVEN does not accept varying time steps, so I simply sample the timeseries at every 6 hours from the starting point. That is what I am doing in the "times" vector. Simply start at 0 (0 hours since the beginning of the forecast) to hour 384 (i.e. 16 days ahead for GEPS) with a step of 6 hours. It's probably not the best option but it works...

Eventually, during maintenance, I'd like to make it so RAVEN is run on the hourly period only (2-3 days), then the final states are saved (the rvc), and then RAVEN is re-run from those states but with the 3-hour data, and so on. So we could maximize the accuracy this way. But that opens another can of worms, regarding hydrograph analysis, graphing, etc.

richardarsenault:
@cjauvin : I've just pushed a commit that does a few things:

1- Added info regarding the start period for the forecasting (see forecasting.py). Basically we need to ensure that we start from the first available timestep in the ECCC forecast. We can control the forecast starting period from Raven directly so let's keep all the info in the netcdf.

2- Added a WPS service for querying and launching realtime (ECCC/Geomet) forecasts using the forecasting package as well ass add the process capabilities in the process _init_.py

3- Added a unit test for this WPS service.

The WPS does not work fully right now, I have a small issue running Raven with the nc_spec functionalities but I suspect it is minor, probably something trivial I forgot, I'll get to that later.

So in summary I've added WPS but this does not change anything from what you've done up till now in the forecasting package.

cjauvin:
@richardarsenault I just pushed a minor update, with which I use the "testdata latest view" url (https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/dodsC/birdhouse/testdata/geps_forecast/GEPS_latest.ncml) and I believe I have now a better understanding of the problem: when the "latest view" points to a `00:00` version of the dataset (e.g. `.../geps/CMC_geps-raw_latlon0p5x0p5_2020120400_allP_allmbrs.nc`), your unit test, in its current version, works. However, when it points to a `12:00` version (e.g. `.../geps/CMC_geps-raw_latlon0p5x0p5_2020120412_allP_allmbrs.nc`), it crashes with an error saying something about "invalid time format used". So our test would crash 50% of the time, 12 hours a day!

richardarsenault:
@cjauvin OK thanks for that update. It's strange, I don't know why that would be the case, since Raven should be able to use the sub-daily forecast times and by using the 0-index time of the netcdf, we are ensuring that we are getting the correct times... so 00:00 or 12:00 should make no difference. There must be something that I am not seeing, maybe @huard or @tlogan2000 can chime in regarding possible issues caused by using the 1200 vs 0000 forecast?

cjauvin:
@richardarsenault Note that it's entirely possible that the error lies in the test itself though, somehow (I don't understand how it works deeply enough currently to be able to assess that).

tlogan2000:
Not sure what is happening but it doesn't seem to be the files themselves. Directly accessing a '00' and '12' netcdf file (not ncml) shows they have identical coordinates except for expected time offset : 

````
url = 'https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/dodsC/birdhouse/eccc/forecasts/geps/CMC_geps-raw_latlon0p5x0p5_2020120400_allP_allmbrs.nc'
url1 = 'https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/dodsC/birdhouse/eccc/forecasts/geps/CMC_geps-raw_latlon0p5x0p5_2020120412_allP_allmbrs.nc'

ds = xr.open_dataset(url, chunks='auto')
ds1 = xr.open_dataset(ur1l, chunks='auto')

for c in ds.coords:
   ...:     print(c, np.unique(ds1[c].values - ds[c].values))

````
Maybe the tests should additionally loop over both a static '00' and '12' as well as the 'current' ncml to make sure things are ok?  I can move these two files back to the thredds 'testdata' folder if needed

**Comment by cjauvin** [comment]
Just to clarify, the `12` netcdf file is having an effect on this particular code:

https://github.com/Ouranosinc/raven/blob/wps_geomet/tests/test_ECCC_forecast.py#L62-L76

It's Raven itself (I believe) that's complaining about some time format issue. This does not happen with the `00` file.

**Comment by tlogan2000** [comment]
@richardarsenault  Not sure this helps but my understanding is that we are trying to fill a gap of a few weeks of availability in the CASPAR data, however caspar data does not seem to ever have the '12' forecast but simply a single file per day?   e.g. https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/catalog/birdhouse/caspar/daily/catalog.html

Also an example file : 'https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/dodsC/birdhouse/caspar/daily/GEPS_20170729.nc'
has 241 time steps  hourly for ~2/3 then jumping to 3h versus our ~96 (3h to 6h)
````
caspar = xr.open_dataset('https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/dodsC/birdhouse/caspar/daily/GEPS_20170729.nc' )
print(caspar.time.diff(dim='time').dt.seconds/3600)
````

Does this have any effect on the RAVEN processes being called?

**Comment by richardarsenault** [comment]
@tlogan2000 maybe there is something related to the 12h set, as yes, for now we've only queried the 0Z forecasts from Caspar. The 12Z forecasts do exist however but I don't think they'll be necessary for hindcasting.

@julemai, is there something you can see that would prevent Raven from running on the 12Z files? maybe the start_date is set to 00:00 and our first timestep is 12h in the forecast file? Just a guess...

**Comment by richardarsenault** [comment]
@huard Could you take a look at the test_wps_realtime_forecast.py please? I am having a problem with raven not finding "nc_spec" for time shifts and deaccumulation flags in this wps. I think I've done everything correctly but I am getting an error: 

AttributeError: No configuration key named nc_spec

I have put a pdb trace just before the offending line to help find the issue, and I've tracked it as best I can but eventually I get lost in the rv and base functions.

**Comment by richardarsenault** [comment]
@cjauvin Not sure where the best place to put this is, but it's a subsetted GEPS 

forecast file that covers the test watershed: (2.4Mb)

[small_GEPS_fcstfile.nc.zip](https://github.com/Ouranosinc/raven/files/5682395/small_GEPS_fcstfile.nc.zip)

File: raven/utilities/forecasting.py [hunk 8]
```diff
@@ -191,6 +191,7 @@ def get_ECCC_dataset(climate_model):
 
     if climate_model == "GEPS":
         file_url = "https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/dodsC/datasets/forecasts/eccc_geps/GEPS_latest.ncml"
+        file_url = "https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/dodsC/birdhouse/testdata/geps_forecast/GEPS_latest.ncml"
```
cjauvin, line 194: Which one is good?

richardarsenault: It's the second one. I left the first one there because eventually that will be the correct path, but for now it's the second one. Perhaps I should have commented it out instead of just overwriting with no info...

Also, I have modified the code a bit to fit both this branch and the hindcasting branch, I'll push something probably in the next hour.

cjauvin: ok sure! if you could comment the first line I think it would make the intention clearer indeed, because to my eyes, it's instantly suspicious (some languages even go as far as having that a syntax error, which I personnaly find is making sense).

richardarsenault: Got it! Will do. Thanks!

It's good that you keep an eye on me, I'm bound to do stupid stuff once in a while 😄

richardarsenault:
@huard @cjauvin In theory, this can be merged, but I think it would be best to delete this one and merge the hindcast_caspar branch instead. 

What I did was merge this branch into hindcast_caspar, and did some editing there, namely to mesh the forecasting codes together and also switch my caspar Process to a RavenProcess like @huard did for the ECCC ones here.

All tests are passing in the other branch, so if my understanding is correct, all changes from here have been included, and improved upon, in the branch "hindcasting_caspar" (#321)

Issue #226, opened 2020-04-14
Split project into ravenpy and the WPS server


huard:
## Description

The python wrapper should be extracted from this repo and live in its own repo. 
This repo would only contain WPS services.

julemai:
Talked to the CSHS folks. We can get for sure a project there: https://github.com/CSHS-CWRA

I will just need to tell them from which GitHub to import. Or if they should just create an empty one with a licence (do you prefer one?) and a name (RavenPy?!).

huard:
An empty one will do. MIT license. ravenpy. And permissions to maintain the repo.

julemai:
I've created the new project `RavenPy` under CSHS. I just became the admin of the whole CSHS GitHub.

I gave @dustming and @richardarsenault write permissions and @huard admin permissions. Let me know whom else you'd like to have added. For now it is private but I can make it also public if you'd like to.

huard:
Thanks! Please make it public, otherwise release could be difficult.

julemai:
Done!

PR #324, opened 2020-11-29
Refactor with RavenPy dependency

cjauvin:
Two things to verify/investigate:

* Does the migration makes sense for all submodules? In particular `utilities`, for which I wasn't 100% sure (maybe the split should be partial to let some stuff in this repo?)
* Some internal tests to `pywps` now seem to fail, not sure why

To keep in mind:
* Although this installs `RavenPy` in editable mode, the binary are still not handled properly in that mode (to fix in RavenPy)


huard (approved):
Looks good ! 

Before merging, I would like the testing infrastructure to be in place for ravenpy. 

So the way I see it:
1. Create raven-testdata
2. Adapt ravenpy to use raven-testdata
3. @tlvu updates the raven test-data PAVICS sync to point to raven-testdata
4. @cjauvin transfers open raven PRs to ravenpy
5. @richardarsenault confirms ravenpy works and tests pass on his end
6. Merge this PR.
7. Purge the raven git history of large binary files

Does that make sense ?

richardarsenault:
@huard that looks like a solid plan to me!

tlvu:
>     7\. Purge the raven git history of large binary files

This is impossible unless the full Raven repo is recreated from scratch, no idea what will happen with old issues/PR.  And the recreated history will be broken since the old commits assume local file that is missing from the recreated repo.

The large files can be deleted by a new commit but will stay in the history forever and forever bloating the history size.

cjauvin:
> The large files can be deleted by a new commit but will stay in the history forever and forever bloating the history size.

Not sure what are the exact implications of using something like this:

https://stackoverflow.com/questions/2100907/how-to-remove-delete-a-large-file-from-commit-history-in-git-repository

huard:
@tlvu I think it is better to render previous versions broken (that are unlikely to ever be reused) than to carry a bloated repo that would hinder further progress and collaborations. 

@cjauvin Yes, BFG was what I was thinking about with this. Ideally it would replace old links with git lfs links, but that doesn't seem to be the case.

tlvu:
> I think it is better to render previous versions broken (that are unlikely to ever be reused) than to carry a bloated repo that would hinder further progress and collaborations.

@huard Old versions (old hash commit id) are also used as reference in PR, CI status, Dockerhub/ReadTheDocs, ... any downstream build/automation activities in general.  Rewriting history means all those references are invalidated.  Other reasons: https://stackoverflow.com/questions/1491001/what-are-the-practical-consequences-of-rewriting-git-history.

`du -sh raven/.git` gives 239M.  It's not too big.  Maybe either leave the history there or cleaner like I did for the birdhouse-deploy repo, create a brand new repo from the last clean commit of the old PAVICS repo so history going forward is clean but old "true history" is still accessible and existing references are not broken.

Of course I leave the final call to you guys since you are the ones interacting day-to-day with this repo so you will feel the impact most.  I simply want to raise awareness of the impact of rewriting git history.

cjauvin:
As I have made progress on the RavenPy packaging and publishing front, I would like to recap on the question of what needs to be ported from the recent changes made in `raven` to `RavenPy`. Can someone please provide a comprehensive checklist?

huard:
Can you just copy over what is in models/ and utilities/  from master ?

cjauvin:
@huard @Zeitsperre @tlvu Here is the current status:

* https://github.com/CSHS-CWRA/RavenPy should be up to date with `raven/master`
* I have made a few recent changes to RavenPy in order for this to work (in particular, I have added this: https://github.com/CSHS-CWRA/RavenPy/blob/master/ravenpy/utilities/testdata.py#L8)
* The test suites for both projects should be all green

tlvu:
>     3\. @tlvu updates the raven test-data PAVICS sync to point to raven-testdata

Done in https://github.com/bird-house/birdhouse-deploy/pull/111

cjauvin:
@huard @Zeitsperre @tlvu Finally the tests are working with RavenPy 0.1.6 pip installed from PyPI.

A couple of notes:
* I have renamed the `raven` CLI tool to `raven-wps` to avoid conflict with the `raven` binary as handled by RavenPy (ideally this whole project would be renamed RavenWPS I guess, to avoid further confusion)
* There is some weird issue with `GDAL_DATA` not pointing in the right place (in the conda env); I had to unset it in my shell before running the tests, to make Fiona happy
* I had to mark as `xfail` two tests in `test_wps_generic_shape_properties.py` (ask @Zeitsperre about it, supposedly it's not related to this PR)

Reminders:
* To install RavenPy with the binaries: `pip install ravenpy --install-option="--with-binaries"`
* To run the tests: `RAVENPY_TESTDATA_PATH=/path/to/raven-testdata pytest`

So from my perspective, the split raven/RavenPy (this PR) is ready to happen, so please tell me if I need to do something more in this context.

huard:
I suggest we go ahead and do the fine-tuning next week.

huard:
Fixes #226

PR #328, opened 2020-12-09
Unlink files that do not have an accompanying md5

### LINKED ISSUE #327: Bug with `get_file`
## Description

`get_file` checks that the checksum matches, but if it doesn't exist, `get_file` will fail the first time (because the md5 check fails) but work the second time since it finds the file in the cache and then doesn't care about the checksum. 

Also I think we should add some doc somewhere about how to contribute new datasets. Ie conventions and rules we follow and how to compute the md5.

Zeitsperre:
## Overview

This PR fixes #327 

Changes:

* Added a check that deletes the downloaded file if an accompanying md5 is not found.

## Additional Information

### DIFF
--- a/raven/tutorial.py
+++ b/raven/tutorial.py
@@ -5,6 +5,7 @@ from typing import Optional, Union
 from pathlib import Path
 from urllib.request import urlretrieve
 from urllib.parse import urljoin
+from urllib.error import HTTPError
 
 from xarray import open_dataset as _open_dataset
 from xarray import Dataset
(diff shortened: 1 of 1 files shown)


huard (approved):

raven/tutorial.py
```diff
@@ -33,9 +34,14 @@ def _get(
...
-        LOGGER.info("Fetching remote file md5: %s" % fullname.as_posix())
-        urlretrieve(url, md5file)
+        try:
+            url = "/".join((github_url, "raw", branch, md5name.as_posix()))
+            LOGGER.info("Fetching remote file md5: %s" % fullname.as_posix())
+            urlretrieve(url, md5file)
+        except HTTPError as e:
+            msg = f"{md5name.as_posix()} not found. Aborting file retrieval."
+            local_file.unlink()
+            raise FileNotFoundError(msg) from e
```
cjauvin, line 44: I was not aware of the `raise .. from ..` pattern.. very interesting subtlety!

PR #329, opened 2020-12-09
ensure branch name is preserved in filepath

### LINKED ISSUE #327: Bug with `get_file`

Zeitsperre:
## Overview

This PR fixes #327 (again)

Changes:

* Added branch name (default: `master`) to the filepath of cached files

## Additional Information


huard (approved):
Works here.

PR #330, opened 2020-12-09
Test era from cache

huard:
## Overview

This PR speeds up the ERA test. It downloads test data from raven-testdata and caches it locally. The test uses the local cache instead of downloading the data every time.


Zeitsperre (approved):
Just did a clean install and ran it. Working perfectly!

huard:
@richardarsenault FYI this speeds up a lot the ERA test. It's not slow anymore.

PR #338, opened 2021-01-20
Fix some notebooks for jenkins automated tests

tlvu:
Trying to get more notebooks pass Jenkins test.

Not all of them are but we have a few less errors now.

It's tested against the latest version of Raven on my devel server https://lvupavics.ouranos.ca/twitcher/ows/proxy/raven/wps.

Jenkins result: http://jenkins.ouranos.ca/job/ouranos-staging/job/lvupavics.ouranos.ca/26/consoleFull

Matching PR to add more testdata: https://github.com/bird-house/birdhouse-deploy/pull/117

@cjauvin FYI we also have a Jenkins that can test the notebooks.


huard:
Changing pavics to boreas is not ideal. I understand how this is easier for testing, but it's not great for the documentation. 
Same thing for datasets display from HTML to text. 
I suggest we have a discussion about our notebook testing strategy.

tlvu:
> Changing pavics to boreas is not ideal. I understand how this is easier for testing, but it's not great for the documentation.

It's to avoid having to replicate all the large `.nc` files that goes with the `.ncml` files to all the test servers.

> Same thing for datasets display from HTML to text.
> I suggest we have a discussion about our notebook testing strategy.

Or we just do not validate those output ... but then we lose the ability to confirm the output matches the server code and testing code.

We could just check the code do not throw exception.  But then we lose half of the advantages of `nbval` which is also to check the output.

I see `nbval` is very similar to the `doctest` module that also checks output.

huard:
Is is because "pavics" is regexped and replaced by the local test machine ? 

Yep, that's the discussion I'd like to have. What role do notebooks play in our integration testing. Do we want to check everything "passes", or yields a value in particular.

tlvu:
> Is is because "pavics" is regexped and replaced by the local test machine ?

Exact.

> Yep, that's the discussion I'd like to have. What role do notebooks play in our integration testing. Do we want to check everything "passes", or yields a value in particular.

I am leveraging tutorials notebooks as end to end testing as well so we get free test cases and we ensure the tutorials are always up-to-date with the code.

Otherwise we need to spend time writing different end to end test cases and still somehow ensure the tutorials are valid as the code changes.

Basically it looks like friction to test the notebooks but it actually save us work in the grand scheme of things.

I think we just have to be more careful in selecting which output is essential and which is just informative.  But I don't think dropping all output checking is the way to go.  Basically when writing and updating notebooks, we should always keep the automated testing side of the notebooks in mind.

For "pavics" vs "boreas", it also boils down to how strict we want the tutorial notebook to be consistent for the user.  I can change that we only "inject" the new hostname for `WPS_URL` variable and so when we print out the `wps_output` url in the output it might not match.  That the original reason why we regex replace all occurences of `pavics.ouranos.ca`.  Hence I had to replace `pavics` with `boreas` for the links I do not want to regex replace.

huard:
My sense is that the main role of notebooks is documentation, and testing is an added bonus. Now I'm getting the feeling that we are making compromises on the documentation to get our tests to run.

tlvu:
The documentation gets tested as well so the bonus goes both ways.  We just need to find a happy middle ground.

tlvu:
@huard Can we merge this PR and then open a new one to relax notebook checking?  It would help testing this other PR https://github.com/Ouranosinc/raven/pull/340 with Jenkins.

huard:
I'm not comfortable with the boreas/pavics change, so I'm ok with merging if we roll that back later.

huard (approved):

PR #339, opened 2021-01-20
Make docker-compose run tests generic and reproducible

tlvu:
Very small PR, read the 2 short commit descriptions.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,7 +18,7 @@ services:
     build: .
     image: raven-wps-tests
     volumes:
-      - /home/christian/gh/raven-testdata:/testdata
+      - ../raven-testdata:/testdata
       - ./:/opt/wps
     environment:
       - RAVENPY_TESTDATA_PATH=/testdata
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -14,11 +14,13 @@ services:
   # (1) Make sure you have raven-testdata locally (https://github.com/Ouranosinc/raven-testdata)
   # (2) Set the left part (local path) in the `volumes` section below to the right path
   # (3) Run: docker-compose run tests
+  #
+  # In case of weird error, rebuild from scratch: docker-compose build --no-cache tests
```
huard, line 18: This does not run the tests on my end. I needed to do `docker-compose run tests` after the build.

huard (approved):

docker-compose.yml
tlvu: Yes the command is just to rebuild.  Some context, the rebuild is done automatically by `docker-compose run tests` only if the image do not exist.  But if the image already exist from a previous run, it should still be fine since we change dependencies rarely and we volume mount in the raven repo so the code is always fresh.  But weird error will happen when the dependencies are out-of-date (ex: xclim too old).

cjauvin: oui en fait je le dis à l'étape (3) juste en haut, mais j'avoue que c'est un peu boiteux mon affaire ..

tlvu: Pushed new commit with some clarification a385439

PR #363, opened 2021-02-03
Update for release 0.11.1 correctly after rush

tlvu:
Ensure `CHANGES.rst` is updated.  We should not rely on Github to keep our release notes.  We might not use Github in the future.


cjauvin (approved):

huard (approved):
Thanks !

PR #369, opened 2021-03-26
Use the zeitsperre anaconda source for the ravenpy dependencies

Zeitsperre:
## Overview

This PR addresses the docker image build issues by relying on the conda-sourced binaries for `ravenc`, `ostrich`, and `RavenPy`.

Changes:

* Simplified the dockerfile install steps for `raven-wps`
* Now using the binaries for `ravenc`, `ostrich`, and `RavenPy` currently hosted on `zeitsperre`

## Additional Information

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ WORKDIR /opt/wps
 # (2) Install RavenWPS in editable mode
 # Have to uninstall the ravenpy installed by conda so the re-install with
 # binaries work.  Same problem as in the Makefile.
-RUN ["/bin/bash", "-c", "source activate wps && pip uninstall --yes ravenpy && pip install ravenpy --install-option=\"--with-binaries\" && pip install -e ."]
+RUN ["/bin/bash", "-c", "source activate wps && pip install -e ."]
 
 # Start WPS service on port 9099 on 0.0.0.0
 EXPOSE 9099
--- a/environment.yml
+++ b/environment.yml
@@ -1,10 +1,12 @@
 name: raven
 channels:
+- zeitsperre
 - conda-forge
 - defaults
 dependencies:
 - pip
 - six
+- ravenpy==0.3.0
 - pywps==4.4.1
 - unidecode
 - birdy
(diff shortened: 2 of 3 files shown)


huard (approved):
Nice ! I guess the plan is to move from a personal channel to a more official channel when things have matured ?

Zeitsperre:
> Nice ! I guess the plan is to move from a personal channel to a more official channel when things have matured ?

Thanks!

That's the plan, from where I see it. Like I've said in the slack, there's an optimization problem (specifically, dependency solving at install takes a long time on some machines, possible because of the `libnetcdf` library that `ravenC` needs) that I'd like to fix down the line (was thinking of reaching out to the RSE slack).

Besides that, we just need a permanent location for the `ravenC` source code (`ostrich` is already using tagged versions on GitHub). If we do throw these binaries on `conda-forge`, we'll need a unique name for `raven` (it's [fairly popular](https://anaconda.org/search?q=raven)).

The recipes seem to be working well enough for now, so I'm glad.

Dockerfile
```diff
@@ -24,7 +24,7 @@ WORKDIR /opt/wps
 # (2) Install RavenWPS in editable mode
 # Have to uninstall the ravenpy installed by conda so the re-install with
 # binaries work.  Same problem as in the Makefile.
```
tlvu, line 26: Sweet !!!  By the way, Makefile also need to undo the `pip install ravenpy --install-option="--with-binaries"`

Zeitsperre: All good. Once It looks like it's running on dockerhub, I'll push a hotfix.

Zeitsperre: Actually, now that I think of it. This is still the non-conda way of installing/running `ravenpy` so it's not technically obsolete. Are we abandoning pure python approaches completely then?

tlvu: I guess we'll still have to keep the pip way but this specific pip way is broken so that's why you guys had to do it the conda way.  So the current pip hack in Makefile is broken ... for the Makefile.

Zeitsperre: It's not broken insofar as it's a bit esoteric/brittle. I'll push a fix for this in the coming week, before we tag a new version. Good eye.

PR #377, opened 2021-04-14
Use ProcessError in RavenProcess._handler

cjauvin:
Following up from https://github.com/CSHS-CWRA/RavenPy/pull/98, add a friendlier error message to the server response (if you accept the fact that an entire stacktrace is something friendly).

### DIFF
--- a/raven/processes/wps_raven.py
+++ b/raven/processes/wps_raven.py
@@ -1,10 +1,13 @@
 import json
 import logging
+import string
+import traceback
 from collections import defaultdict
 from pathlib import Path
 
 from pywps import Format, LiteralOutput, Process
-from ravenpy.models import Raven
+from pywps.app.exceptions import ProcessError
+from ravenpy.models import Raven, RavenError
 from ravenpy.models.rv import HRU
 from ravenpy.utilities.checks import single_file_check
 from ravenpy.utilities.io import archive_sniffer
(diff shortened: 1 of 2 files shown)


huard (approved):

raven/processes/wps_raven.py
```diff
@@ -144,7 +147,15 @@ def _handler(self, request, response):
...
 
         # Launch model with input files
-        self.run(model, ts, kwds)
+        try:
+            self.run(model, ts, kwds)
+        except Exception as exc:
+            err_msg = traceback.format_exc()
+            # By default the error message is limited to 300 chars and strips
+            # many special characters
+            raise ProcessError(
```
cjauvin, line 156: @huard @Zeitsperre Another possibility here would be to test for a `RavenError` here and only return its message, in this context.

huard: I think the traceback is useful, at least while we're still moving things around a lot.

tlvu: Sorry stupid question but what would happen before if `self.run(model, ts, kwds)` throws an exception? Where will it go?  Why do we have to catch and raise again?

tlvu (approved):
Approve the spirit of this PR.  Dont know the code enough to review the code.

raven/processes/wps_raven.py
huard: To add the traceback to the error message, and return the standard pywps ProcessError exception.

huard:
We've got errors due to the libnetcdf version... ? what's up with that ?

PR #382, opened 2021-04-16
update remaining nb. mv bias to sandbox

huard:
## Overview

This PR fixes Finch related notebooks. 

Note that some processes only work with GR4JCN because other emulators do not yet support the area, lat, lon, elevation -> hru conversion.


.cruft.json
```diff
@@ -23,7 +23,7 @@
             "project_repo_name": "raven",
             "project_readthedocs_name": "pavics-raven",
             "project_short_description": "Raven offers processes related to hydrological modeling, and in particular, the Raven hydrological modeling framework.",
-            "version": "0.12.0",
+            "version": "0.12.1-beta",
```
tlvu, line 26: bump again to make it without "beta"?

huard: Ah yes, darn. Hate that. I'll wait for the tests to complete. Bump it then merge without waiting for the tests.

huard: The docs failed for some reason, but I'll ignore it for now.

PR #389, opened 2021-05-20
notebooks: refresh output for Jenkins

tlvu:
## Overview

Trying to get Raven notebooks to pass Jenkins without `--nbval-lax` since we have disabled by default comparing html and javascript outputs, see PR https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/73

Changes:

* A run of `make refresh-notebooks`


Zeitsperre:
@tlvu I think this should be the last PR before we rename Raven -> RavenWPS. Think you could verify that the "refreshed" notebooks are working on Jenkins?

huard:
Is there any actual code change here or this is just a refresh ?

tlvu:
Pretty sure I only did a refresh, no code change.  That's why it was in draft mode.  I was not fully ready to merge it yet.  I intended to make a few code change to get Jenkins to pass.  Just ignore this for now if you need to make code change.  I'll refresh again later when I can continue on this.

PR #393, opened 2021-06-10
Extract_geographical_watershed_properties.ipynb: fix run failure in new Jupyter env

tlvu:
Fixes https://github.com/Ouranosinc/raven/issues/392 using hints from https://github.com/Ouranosinc/raven/issues/388. Closes #388.

Note I am not sure what I am doing as I do not understand the code.  I simply try to get around the errors and produce similar graph.  The old graph had `spatial_ref` in the title, the new one don't.  Wonder if it's something critical.

Clarification from Trevor: 

 ==========
Putting it simply, the object returned (`xarray.DataArray`) is different from what the following cell needed to handle for (`rasterio._io.RasterReader` → `numpy.ndarray`):
```
23:00:58  Input:
23:00:58  # Note that geojson, and rasterio/rioxarray need to be installed for this to work.
23:00:58  # In the next birdy release, the tiff grid will be automatically converted to a DataArray.
23:00:58  # Here we need to manually convert it.
23:00:58  features, statistics, grid = stats_resp.get(asobj=True)
23:00:58
23:00:58  with rasterio.io.MemoryFile(grid[0]) as memfile:
23:00:58      with memfile.open() as dataset:
23:00:58          grid = rio.open_rasterio(dataset)
23:00:58
23:00:58  grid.plot()
```
The `spatial_ref`title is something that `rasterio` adds to the plot, and it isn't at all needed. The notebook should work perfectly so long as the client running it has `rioxarray`available in their ipython kernel.

 ==========

With the new Jupypter env for latest RavenPy `0.5.2` from PR https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/75 `Extract_geographical_watershed_properties.ipynb` fails with the following error below.

```
23:00:58  _ raven-master/docs/source/notebooks/Extract_geographical_watershed_properties.ipynb::Cell 7 _
23:00:58  Notebook cell execution failed
23:00:58  Cell 7: Cell execution caused an exception
23:00:58
23:00:58  Input:
23:00:58  # Note that geojson, and rasterio/rioxarray need to be installed for this to work.
23:00:58  # In the next birdy release, the tiff grid will be automatically converted to a DataArray.
23:00:58  # Here we need to manually convert it.
23:00:58  features, statistics, grid = stats_resp.get(asobj=True)
23:00:58
23:00:58  with rasterio.io.MemoryFile(grid[0]) as memfile:
23:00:58      with memfile.open() as dataset:
23:00:58          grid = rio.open_rasterio(dataset)
[...]

PR #395, opened 2021-06-16
Fix broken refresh target

tlvu:
## Overview

Fix the wrong output file path and not replacing the output url properly with the refresh target.  See commit description for details.

## Related Issue / Discussion

https://github.com/Ouranosinc/raven/pull/393#issuecomment-860952912

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ WPS_PORT := 9099
 WPS_URL := http://0.0.0.0:$(WPS_PORT)
 
 # If WPS_URL is overridden, this should also be overridden to match.
-WPS_OUTPUT_URL := http://localhost:$(WPS_PORT)
+WPS_OUTPUT_URL := http://localhost:$(WPS_PORT)/outputs
 
 GDAL_VERSION := $(shell gdal-config --version)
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -244,7 +248,7 @@ refresh-notebooks:
 NB_REFRESH_FILE := ""
 .PHONY: refresh-notebooks-impl
 refresh-notebooks-impl:
-	bash -c 'WPS_URL="$(WPS_URL)" FINCH_WPS_URL="$(FINCH_WPS_URL)" FLYINGPIGEON_WPS_URL="$(FLYINGPIGEON_WPS_URL)" jupyter nbconvert --to notebook --execute --ExecutePreprocessor.timeout=240 --output "$(NB_REFRESH_FILE)" "$(NB_REFRESH_FILE)"; sed -i "s@$(WPS_URL)/outputs/@$(OUTPUT_URL)/@g" "$(NB_REFRESH_FILE)"'
+	bash -c 'WPS_URL="$(WPS_URL)" FINCH_WPS_URL="$(FINCH_WPS_URL)" FLYINGPIGEON_WPS_URL="$(FLYINGPIGEON_WPS_URL)" jupyter nbconvert --to notebook --execute --ExecutePreprocessor.timeout=240 --output "$(NB_REFRESH_FILE)" --output-dir . "$(NB_REFRESH_FILE)"; sed -i "s@$(WPS_OUTPUT_URL)/@$(OUTPUT_URL)/@g" "$(NB_REFRESH_FILE)"'
```
Zeitsperre, line 251: Won't changing the output dir to `.` place the notebook at the location where this command is called? Why not use `in-place`? See: https://nbconvert.readthedocs.io/en/latest/usage.html#notebook-and-preprocessors

tlvu: Oh funny, I didn't notice that `--inplace` option yesterday with `jupyter nbconvert --help`.  However setting `--output "$(NB_REFRESH_FILE)" --output-dir .` does exactly the same thing since `$(NB_REFRESH_FILE)` is the full path of the notebook.

Zeitsperre (approved):

PR #397, opened 2021-07-09
Support WFS point filter from RavenPy [WIP]

Zeitsperre:
… RavenPy with point intersect

## Overview

This PR is an intermediary branch. Configurations will need to be updated to published packages before merging.

Changes:

* Updated dependencies to be synchronous with RavenPy
* Temporarily moved dependencies to dev branches for testing purposes

## Related Issue / Discussion

I have been getting segfaults for some strange reason (see: https://gist.github.com/Zeitsperre/d5d5f88be82835a47e48172f3bb7be16)

## Additional Information

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -8,15 +8,15 @@ dependencies:
 - ravenpy>=0.5.1
 - pywps==4.4.2
 - unidecode
-- birdy
-- owslib
+#- birdy
+#- owslib
 - poppler
 - jinja2
 - click
 - psutil
 - numpy
 - scipy
-- xarray==0.17
+- xarray
 - xclim
 - rioxarray
 - matplotlib
@@ -46,3 +46,6 @@ dependencies:
 - pip:
   - spotpy
   - urlpath
+  - owslib @ git+https://github.com/geopython/OWSLib.git@master
+  - birdhouse-birdy
+  - ravenpy @ git+https://github.com/CSHS-CWRA/RavenPy.git@add_point_intersection
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ endif
 
 CONDA := $(shell command -v conda 2> /dev/null)
 CONDA_ENV ?= $(APP_NAME)
-PYTHON_VERSION = 3.6
+PYTHON_VERSION = 3.7
 
 # Choose Anaconda installer depending on your OS
 ANACONDA_URL = https://repo.anaconda.com/miniconda
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -17,6 +17,7 @@ dependencies:
 - bumpversion
 - twine
 - cruft
+- pre-commit
 - pytest-dependency
 - pytest-notebook
 - sphinx-autoapi
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,8 @@
 ravenpy>=0.4.2
 six
 pywps==4.4.2
-owslib
+# owslib
+owslib @ git+https://github.com/geopython/OWSLib.git@master
 lxml
 jinja2
 click


huard (approved):

huard:
Are there notebooks to fix as well ?

tlvu:
Would this PR unblock upgrading geoserver (PR https://github.com/bird-house/birdhouse-deploy/pull/136)?

Zeitsperre:
@huard This PR is far from ready. I keep getting segfaults on GitHub and locally. Investigating further today.

Zeitsperre:
There's something up with GeoPandas in Raven-WPS:

- Reading a feature converted to JSON from byte stream gives me segfaults. Writing out that object and opening with GeoPandas in a fresh env? No problem. 
- Reading the same object and opening in the RavenWPS-installed GeoPandas. Segfault.

Something isn't adding up.

Zeitsperre:
Fresh env versions:
- Pandas@1.3.0
- GeoPandas@0.9.0
- PyProj@3.1.0
- Proj@8.0.1
- Geojson@2.5.0
- GDAL@3.3.1

RavenWPS env versions:
- Pandas@1.2.5
- GeoPandas@0.9.0
- PyProj@3.1.0
- Proj@7.2.0
- Geojson@2.5.0
- GDAL@3.1.4

Zeitsperre:
I think I've found the source of the segfault using gdb. It's definitely GeoPandas/GEOS related:
```
Thread 1 "python" received signal SIGSEGV, Segmentation fault.
0x00007fff9e26275e in get_geom () from /home/tjs/miniconda3/envs/raven-wps/lib/libesmf_fullylinked.so
```
Will look further into this tomorrow.

Zeitsperre:
Confirmed: It looks like `xesmf` or (more likely) `esmf` are the culprits here. Removing these libraries fixes the issues. Here's the working env for reference: https://gist.github.com/Zeitsperre/e31c3f11b745856c655157d38ef4ec42

@huard I'm not sure how to proceed here. I'm not even sure where the problem is beyond my segfault readout.

Zeitsperre:
Everything seems to be working, barring a few tests on the actual RavenPy/Raven-Hydro modelling side:
- TestOstrichGR4JCemaNeigeProcess.test_simple
```
  File "/home/tjs/git/raven/raven/processes/wps_raven.py", line 138, in _handler
    self.run(model, ts, kwds)
  File "/home/tjs/git/raven/raven/processes/wps_raven.py", line 112, in run
    model(ts=ts, **kwds)
  File "/home/tjs/git/RavenPy/ravenpy/models/base.py", line 404, in __call__
    procs = self.run(ts, overwrite, **kwds)
  File "/home/tjs/git/RavenPy/ravenpy/models/base.py", line 387, in run
    cmd = self.setup_model_run(tuple(map(Path, ts)))
  File "/home/tjs/git/RavenPy/ravenpy/models/base.py", line 258, in setup_model_run
    self.derived_parameters()
  File "/home/tjs/git/RavenPy/ravenpy/models/emulators/gr4jcn.py", line 190, in derived_parameters
    one_minus_CEMANEIGE_X2=(1.0 - params.CEMANEIGE_X2)
AttributeError: 'NoneType' object has no attribute 'CEMANEIGE_X2'
```
- TestOstrichMOHYSEProcess.test_simple
```
  File "/home/tjs/git/raven/raven/processes/wps_raven.py", line 138, in _handler
    self.run(model, ts, kwds)
  File "/home/tjs/git/raven/raven/processes/wps_raven.py", line 112, in run
    model(ts=ts, **kwds)
  File "/home/tjs/git/RavenPy/ravenpy/models/base.py", line 404, in __call__
    procs = self.run(ts, overwrite, **kwds)
  File "/home/tjs/git/RavenPy/ravenpy/models/base.py", line 387, in run
    cmd = self.setup_model_run(tuple(map(Path, ts)))
  File "/home/tjs/git/RavenPy/ravenpy/models/base.py", line 258, in setup_model_run
    self.derived_parameters()
  File "/home/tjs/git/RavenPy/ravenpy/models/emulators/mohyse.py", line 195, in derived_parameters
    par_rezi_x10 = 1.0 / params.par_x10
AttributeError: 'NoneType' object has no attribute 'par_x10'
```
Beyond that, removing clisops is a bit tricky; We currently bundle it in the xclim conda package therefore, for the time being, I've set xclim to install from PyPI. It would be worthwhile to perhaps decouple these, especially since clisops has gotten much more all-encompassing (regridding, bias correction).

@cjauvin If this PR looks good, I'll see if I can remove as many git installs as possible before merging. If you want to add fixes for the breaking tests, feel free to push here or base a new branch off this one. Your call.

cjauvin:
@Zeitsperre My last commit should fix your two errors above, but the fix should be validated ideally by @huard and @julemai because to be honest I don't even understand how it could work before.

Zeitsperre:
This PR is waiting on the following:

1. A new `ravenpy` version based off the current master branch (v0.7.0)
2. A new version/build of `xclim` that does not bundle `clisops` (https://github.com/Ouranosinc/xclim-feedstock/issues/1)
3. Verification and approval (verify that the default parameter values added are sane and that the updated WFS functions are acceptable).

Thanks for the help @cjauvin !

cjauvin:
I just released RavenPy 0.7.0 on PyPI.

Zeitsperre:
Once xclim 0.28.1 is in the conda-forge index (tomorrow, most likely), everything should work as intended.

environment.yml
```diff
@@ -16,15 +18,15 @@ dependencies:
...
 - scipy
-- xarray==0.17
-- xclim
+- xarray >=0.18
+- xclim >=0.28.1
 - rioxarray
 - matplotlib
 - dask
-- pandas
+- pandas <1.3
```
huard, line 26: Is this still necessary ?

raven/processes/wps_ostrich_gr4j_cemaneige.py
```diff
@@ -44,6 +44,15 @@
...
 
+params = LiteralInput(
+    "params",
+    "Comma separated list of model parameters",
+    abstract="UParameters: " + ", ".join(f.name for f in fields(params_defaults)),
+    data_type="string",
+    default=", ".join(map(str, astuple(params_defaults))),
+    min_occurs=0,
+)
+
```
huard, line 55: Why is this needed ?

raven/processes/wps_ostrich_gr4j_cemaneige.py
```diff
@@ -78,11 +87,16 @@ class OstrichGR4JCemaNeigeProcess(OstrichProcess):
     title = ""
     version = ""
     model_cls = GR4JCN_OST
-    tuple_inputs = {"lowerBounds": GR4JCN_OST.Params, "upperBounds": GR4JCN_OST.Params}
+    tuple_inputs = {
```
huard, line 90: Same.

raven/processes/wps_ostrich_gr4j_cemaneige.py
```diff
@@ -78,11 +87,16 @@ class OstrichGR4JCemaNeigeProcess(OstrichProcess):
...
+        "params": GR4JCN_OST.Params,
+        "lowerBounds": GR4JCN_OST.Params,
+        "upperBounds": GR4JCN_OST.Params,
+    }
     inputs = [
         wio.ts,
         wio.nc_spec,
         wio.nc_index,
+        params,
```
huard, line 99: Same.

huard (commented):
Why did you need to add params to the ostrich processes ?

environment.yml
Zeitsperre: No, the bug has been resolved. Will modify.

raven/processes/wps_ostrich_gr4j_cemaneige.py
Zeitsperre: Difficult to determine from my side. As far as I can tell, my changes don't touch these processes, but still the tests were failing (see my comment above).

huard: @cjauvin any idea why the ravenpy update would require params to be defined in ostrich ?

cjauvin: I'm not entirely sure but I think it's due to that fact that some emulators (`GR4JCN` in particular) do not override the `derived_parameters` method for their Ostrich counterpart (i.e. `GR4JCN_OST`) which means that the one from parent class is called (`GR4JCN.derived_parameters`), and `self.config.rvp.params` is expected to be defined in it:

https://github.com/CSHS-CWRA/RavenPy/blob/8c5f9ba7792f53821ca96fc0415d72b217180764/ravenpy/models/emulators/hbvec.py#L236-L257

huard: But I'm not sure I understand why derived_parameters was removed in 
https://github.com/CSHS-CWRA/RavenPy/commit/59cc90e895a21233d1ff16abe0b000fdbf9cd58a#diff-9fe4fb4dc82db51bc5391ff9a00afacc03509f2c3a1bdf1b0619b41d7946b8ed

Should I just add it back, or is there a new framework I should work with ?

PR #403, opened 2021-08-30
Pin Xclim below 0.29.0

Zeitsperre:
Xclim 0.29.0 introduced a breaking API change for SDBA objects. This temporarily pins xclim below this change.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -19,7 +19,7 @@ dependencies:
 - numpy
 - scipy
 - xarray >=0.18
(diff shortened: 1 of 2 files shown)


Zeitsperre:
Using newest ravenpy here, as the xclim version is already pinned there.

tlvu (approved):
LGTM.  Great that RavenPy already pin Xclim so I won't have to pin Xclim myself when I'll rebuild the Jupyter env.

PR #408, opened 2021-09-10
update to ravenpy 0.7.5

### LINKED ISSUE #407: Update notebooks to xclim 0.29 using new sdba API

huard:
## Overview

This PR fixes #407 

Changes:

* Compatible with xclim 0.29
* Update notebooks
* Remove FLYINGPIGEON from Makefile
* Do not fail the doc build on warnings
* Fix some of the docs

## Additional Information


Zeitsperre:
@huard 

Good news: Looks like the test has nothing to do with Raven or RavenPy. Bad news: Looks like the test has nothing to do with Raven or RavenPy.

@cjauvin and I are going to look into trying to reproduce this segfault (I haven't had luck on my machine natively; I'll be testing with a VM tomorrow). If we can get a core dump, we should be able to glean where the problem is. Christian has a suspicion that it might have something to do with the NetCDF4 C-library. I think it might have something to do with climpred and xclim.

cjauvin:
Mauvaise nouvelle, je viens d'essayer dans un environnement frais et ça ne plante pas, sur ma machine Ubuntu.

cjauvin:
Ah non je me reprends! Le test individuel (`/test_wps_climpred_hindcast_verification.py`) ne plante pas quand on l'appelle seul, mais il plante (segfault) quand on appelle la suite de tests au complet.

Je me sousviens d'avoir eu des problèmes de ce genre y a qq mois, c'est quasiment impossible à déboguer.

huard:
Je pense avoir déjà utilisé une approche de style bissection pour comprendre ce qui se passe. Il y a une combinaison de tests qui cause le problème, il faut trouver l'autre.

Zeitsperre:
Looks like @cjauvin is onto something with his NetCDF4 theory - It's traceable to the HDF5 source library: https://gist.github.com/Zeitsperre/150dbed5b2772938e36bd00dea25ddef

Zeitsperre:
It seems that the two tests that cause the segfault are `test_wps_climatology_esp()` and `test_wps_climpre_hindcast_verification()`. It could be that there are too many threads called HDF5 (asynchronous file reads suing dask + HDF5?) being created during `test_wps_climatology_esp()` that aren't being cleared from the memory heap fast enough before the next test runs. For perspective: 
image

Possibly related issue with HDF5: https://forum.hdfgroup.org/t/heap-use-after-free-by-the-call-h5aread-when-running-with-multiple-threads-in-hdf5-thread-safe-version/6405

Zeitsperre:
@huard Just to be clear, the climpred test is being skipped currently because of the segfault. Are we merging this PR with this test disabled?

huard:
I suggest we do that, but please open an issue to re-enable the test.

cjauvin:
Le mieux serait d'essayer d'isoler et reproduire le problème dans la composante la plus low-level possible, pour ouvrir une issue dans la librairie concernée, quelle qu'elle soit.

Zeitsperre (approved):
Approving for now, but we need to prioritize resolving #412

huard:
Ou de trouver une combinaison de versions qui ne pose pas de problème.

PR #416, opened 2021-12-15
Update docs

### LINKED ISSUE #356: Notebook Bias_correcting_climate_data produces impossible result
Notebook "Bias_correcting_climate_data" works if we set the times to 12h instead of 00h (i.e. 2070,1,1,12 to 2079,12,31,12), but the results are very wrong, with essentially 0 flow for the last 9 out of 10 years for a ~4500km2 catchment. 

We would need to check what is happening that is causing the flow to be essentially 0 for 10 years.

richardarsenault:
## Overview

This PR adds the tutorial notebooks to the documentation. Added information and fixed notebook titles to appear nicely in the docs. 

Also deletes the unnecessary bias correction notebook that was a WIP, a new cleaner version has been provided in the tutorial section.

Fixes #356.


Zeitsperre (approved):
LGTM

PR #453, opened 2023-01-16
Correct datetime for some models

Mayetea:
In the tutorial 08_Getting_and_bias_correcting_CMIP6_data, the datetime format of some models weren't in the good format for the data extraction of the desired duration. I've managed to correct the issue by transforming the non-standard calendar in numpy.datetime64 which is the desired format. It is now possible to use these models also: CMCC-ESM2, CanESM5, FGOALS-g3, INM-CM4-8 and INM-CM5-0.


huard:
I suggest using `ds = ds.convert_calendar('standard')` instead. 

See https://docs.xarray.dev/en/stable/generated/xarray.Dataset.convert_calendar.html

Mayetea:
I've changed the way we convert the calendar and it's working. I pushed it, so a new review would be appreciated. Thanks!

Zeitsperre:
To give this a chance to get into the next version, I'll wait for @huard's review before preparing the release.

huard:
There are a couple of weird warnings in the nb, can these be fixed ?
Also, the last graphics at the end of the nb seem to be missing.

Mayetea:
I have the graphics on my side. I'm looking into the odd, Latitude warning. The rest seems to be good

Mayetea:
So the warning for Latitude were from the chosen model, which crashes in the original nb. I switch back to MIROC6 (like in the original) since it doesnt have any odd warning. The other warnings were already in the original nb, but I can look into it in another issue/PR.

Zeitsperre (approved):

PR #461, opened 2023-02-22
 Use RavenPy v0.11.0

Zeitsperre:
## Overview

Changes:

* Updated several dependencies to be in sync with `ravenpy` v0.11.0 (including `raven-hydro` v3.6)
* Updated pre-commit hooks
* Modified (workaround) some processes to reflect breaking changes in `fiona` v1.9
   * places upper limit below v2.0
* Updated default `pytest-xdist` scheduler to use `worksteal`
* Added `MAIN_TESTDATA_BRANCH` as a modifiable env var in main CI workflow

## Related Issue / Discussion

The changes in fiona are breaking to `python-rasterstats`. I needed to hack a workaround for fixing the properties out. Much of this logic should be removed when a new version is released. Zonal statistics processes need to be rewritten from the ground up anyway.

Other changes stemming from the new fiona popped up in a few other places as well. GIS function in RavenPy need to be torn apart and re-written. Good motivation to continue work on RavenGIS.

## Additional Information

https://github.com/perrygeo/python-rasterstats/issues/281

* **The testing branch of raven-testdata must be modified back to `master` before merging** (see: https://github.com/Ouranosinc/raven-testdata/pull/34)
   * Done.

PR #462, opened 2023-02-22
use provision-with-micromamba action

Zeitsperre:
## Changes

* Replaced `setup-miniconda` action with `provision-with-micromamba` action, using caching


Zeitsperre:
This will be safe to merge after #461

huard (approved):

PR #463, opened 2023-02-28
Prepare v0.17.0

Zeitsperre:
## Overview

Changes:

* Updated history of changes and bumped version to v0.17.0


Zeitsperre:
@Mayetea

I'll wait until the end of the day for #453, if that works for you. If you think you need more time, let me know when you can. We can push the release a day or two if needed.

huard (approved):

PR #464, opened 2023-04-03
Delete unnecessary wps

richardarsenault:
## Overview

This is a first pass at removing WPS support for Raven-based modelling. This follows the recent refactoring and migration of modelling capabilities to the Ravenpy package. Notebooks, WPS processes and tests related to raven-based modelling have been deleted. Only those pertaining to geoserver have been kept.

CHANGES.rst has been modified to document this.

Future steps would require modifying the client environment, setup.py, installation packages and other config to ensure RAVEN and OSTRICH binaries are not built for nothing when deploying the service.

This is a first pass to help accelerate the process, but @huard, @Zeitsperre or @tlvu should probably double and triple check to ensure this is both desirable and correctly implemented. If it was poorly executed or unnecessary, simply delete this branch and PR!


huard (commented):
I suggest we keep the notebooks:
- Region_selection
- Subset_climate_data_over_watershed

huard:
# Documentation
- The installation docs referring to the installation of Raven and Ostrich should be deleted. 
- There are no processes in the process page

Zeitsperre:
Looking over these changes, I want to mention the following:
 * The `requirements.txt` and `environment.yml` files need to be updated to include only the libraries that are needed to run the remaining processes (RavenPy will set the underlying process dependencies)
 * One particular dependency (`pssh`) is not listed on conda-forge. From what I can tell, this is a parallel version of `paramiko`. Should I open an issue at https://github.com/ParallelSSH/parallel-ssh to ask for a conda package?

PR #485, opened 2023-09-22
Configurable servers

Zeitsperre:
## Overview

FYI @fmigneault 

Changes:

* Allow for configurable GeoServer
* Remove mentions of flyingpigeon in docs

## Additional Information

- https://github.com/bird-house/birdhouse-deploy/issues/14

## Additional Information (2025-08)

The Dockerfile `CMD` now formats arguments by using shell overrides. This makes it possible to make use of `RAVEN_BIND_HOST` and `RAVEN_BIND_PORT`. It would be helpful if I could get confirmation that this is desired (versus something like `RAVEN_BIND_ADDRESS`, including both `host:port`).

Whatever we choose to employ, it should be coordinated and cover the scope of the active Birdhouse services at the same time. The other features (like having a more configurable GeoServer URL: `RAVEN_GEOSERVER_URL`) are useful/valuable to have.

Once this is released and added to Birdhouse-deploy, deployments will no longer need the `GEO_URL` variable (replaced by `RAVENPY_GEOSERVER_URL` and `RAVEN_GEOSERVER_URL`).


Dockerfile
```diff
@@ -17,16 +17,20 @@ RUN mamba env create -n raven -f environment.yml \
...
 # Install raven
 RUN pip install . --no-deps
 
-# Start WPS service on port 9099 on 0.0.0.0
-EXPOSE 9099
+# Start WPS service on port 9099
+EXPOSE $RAVEN_BIND_PORT
 
-CMD ["gunicorn", "--bind=0.0.0.0:9099", "raven.wsgi:application"]
+CMD ["gunicorn", "--bind=$RAVEN_BIND_ADDRESS:$RAVEN_BIND_PORT", "raven.wsgi:application"]
```
fmigneault, line 33: I am pretty sure `$<env>` format does not work in `CMD`. 
Not sure about `EXPOSE`. 

The workaround is to use an entrypoint shell script, which then uses the variables.

docs/source/dev_guide.rst
```diff
@@ -79,8 +79,7 @@ Do the same as above using the ``Makefile``.
 Running notebooks tests
 -----------------------
 
-Assuming that the ``raven`` conda env has already been created and is up-to-date and
-raven-wps has been installed with ``make develop``::
+Assuming that the ``raven`` conda env has already been created and is up-to-date and `raven-wps`` has been installed with ``make develop``::
```
fmigneault, line 82: Missing a ``` ` ``` before `raven-wps`

Highlighting is better if using the following instead of `::`
```rst
.. code-block:: shell
  
    [...code...]

```

fmigneault (commented):
What would be the combination of env-var and config overrides that you found to fix https://github.com/bird-house/birdhouse-deploy/issues/14 would you recommend? 

A sample `docker run` showing different bind/host/port for certain results would be useful.

Makefile
```diff
@@ -7,19 +7,23 @@ APP_NAME := raven-wps
 
 OS := $(shell uname)
 
-WPS_PORT := 9099
-WPS_URL := http://0.0.0.0:$(WPS_PORT)
+WPS_PORT ?= 9099
+WPS_URL ?= http://0.0.0.0:$(WPS_PORT)
```
tlvu, line 11: Makefile noob question, what's the difference between `:=` and `?=`?  The former `:=` already works fine by the way.

raven/utilities/geoserver.py
```diff
@@ -1,12 +1,11 @@
...
-GeoServer interaction operations.
+"""GeoServer interaction operations.
 
 Working assumptions for this module:
 * Point coordinates are passed as shapely.geometry.Point instances.
 * BBox coordinates are passed as (lon1, lat1, lon2, lat2).
 * Shapes (polygons) are passed as shapely.geometry.shape parsable objects.
 * All functions that require a CRS have a CRS argument with a default set to WGS84.
-* GEO_URL points to the GeoServer instance hosting all files.
+* GEOSERVER_URL points to the GeoServer instance hosting all files.
```
tlvu, line 8: Oh, rename of a public configuration variable name.  Is there a way to still support the old `GEO_URL` var name for backward compatibility?  Any existing deployments, unless the admin read the Changelog, will not know about this variable rename.

Breaking the public API is the main reason we often have to pin older version in our Jupyter env build while waiting for downstream package to adapt to the new API.  Let's not continue this bad habit of breaking backward compatibility.

raven/utilities/geoserver.py
```diff
@@ -46,11 +45,15 @@
...
 
+from .geo import determine_upstream_ids
+
 # Do not remove the trailing / otherwise `urljoin` will remove the geoserver path.
-# Can be set at runtime with `$ env GEO_URL=https://xx.yy.zz/geoserver/ ...`.
-GEO_URL = os.getenv("GEO_URL", "https://pavics.ouranos.ca/geoserver/")
+# Can be set at runtime with `$ env RAVENPY_GEOSERVER_URL=https://xx.yy.zz/geoserver/ ...`.
+GEOSERVER_URL = os.getenv(
+    "RAVENPY_GEOSERVER_URL", "https://pavics.ouranos.ca/geoserver/"
+)
```
tlvu, line 55: One way to keep this backward compat is to keep `GEO_URL = os.getenv("GEO_URL", "https://pavics.ouranos.ca/geoserver/")` and change `GEOSERVER_URL` to `GEOSERVER_URL = os.getenv("RAVENPY_GEOSERVER_URL", GEO_URL)`

Dockerfile
tlvu: One easy way to confirm env var works in `CMD` and `EXPOSE` is to build the docker image and test it in the PAVICS stack.

tlvu (changes requested):
Cool change, thanks.  I simply would like to have backward compat with `GEO_URL` and ensure the new Docker image works.

Makefile
fmigneault: `:=` sets the value, `?=` sets the value if not already set (good for defaults)

There's also a `=` that does the same as `:=`, but allows dynamic resolution of the right part only when referenced somewhere else by `$( )` instead of on the line the `=` is indicated. This allows, for example, computing a `$(WPS_PORT)` value later on, and still obtain the desired result of the "final" value.

Dockerfile
fmigneault: I did that and as I remembered, docker run raises something related to `$RAVEN_BIND_ADDRESS:$RAVEN_BIND_PORT` being invalid because `$RAVEN_BIND_PORT` doesn't get replaced by the provide integer.

raven/utilities/geoserver.py
Zeitsperre: Done.

Dockerfile
tlvu: Given `ENV RAVEN_BIND_PORT=${RAVEN_BIND_PORT:-9099}`, if we can make the launching command "see" the env var, it should work.  So probably use `ENTRYPOINT` instead of `CMD`.  If needed, force starting a new shell to "see" the env var with `bash -c "gunicorn --bind=$RAVEN_BIND_ADDRESS:$RAVEN_BIND_PORT raven.wsgi:application"`.

Or research a recipe for how to properly start gunicorn inside a Docker image.

huard (approved):

tlvu (approved):
LGTM

Zeitsperre:
@fmigneault I'm curious to hear your thoughts on the recent changes to the Dockerfile here. Does the `sh`-approach work for us, or does that open us up to new problems (logistics/security)?

Dockerfile
fmigneault: I think this could be a viable workaround. I don't know if there is any impact though of running `gunicorn` via a sh command (eg: limitations in child/subprocesses/threads)? Probably negligible.

docs/source/dev_guide.rst
```diff
@@ -125,20 +121,19 @@ If instead we want to run the notebooks against the production raven-wps server
...
-Or:
+    Or, alternatively:
 
     .. code-block:: console
 
         # to test a single notebook (note the .run at the end of the notebook path)
         make WPS_URL=https://pavics.ouranos.ca/twitcher/ows/proxy/raven/wps docs/source/notebooks/Subset_climate_data_over_watershed.ipynb.run
 
-We can also override all three of the server variables (WPS_URL, FINCH_WPS_URL, FLYINGPIGEON_WPS_URL) to pick and choose any servers/services from anywhere we want.
+We can also override all three of the server variables (``WPS_URL``, ``FINCH_WPS_URL``) to pick and choose any servers/services from anywhere we want.
```
tlvu, line 131:
```suggestion
We can also override both of the server variables (``WPS_URL``, ``FINCH_WPS_URL``) to pick and choose any servers/services from anywhere we want.
```

tlvu (approved):
LGTM.  Did you build locally a docker image then deploy it on your PAVICS server and target it with Jenkins?

fmigneault:
@tlvu 
You're right. It is I that misinterpreted the diff. Since the option was already there, I am also less worried about it.

fmigneault (approved):

Zeitsperre:
Thanks everyone for your help with this. Merging now!

Issue #494, opened 2024-05-07
Licensing violations


Zeitsperre:
Briefly looking at some of the build failures, it occurred to me that a few of our dependencies are licensed for GPLv2-only and GPLv3-only. This isn't good.

Problematic packages:
- pysheds (GPLv3)
- unidecode (GPLv3)

Fixing this may mean that we pull a few processes. Hopefully not much.

Zeitsperre:
We might be able to substitute [anyascii](https://github.com/anyascii/anyascii) for `unidecode`. Worth exploring.

PR #495, opened 2024-05-13
Drop pysheds

Zeitsperre:
## Overview

This PR addresses points raised in #494 

Changes:

* Removed `pysheds` due to a licensing violation

## Related Issue / Discussion

https://www.apache.org/licenses/GPL-compatibility.html


Zeitsperre:
It has to do with `pyogrio`. I may need to pin that at or below `0.7.4`.

PR #505, opened 2024-05-13
Remove unidecode (for anyascii), pandoc, poppler, and tbb

Zeitsperre:
## Overview

This PR fixes [issue id]

Changes:

* Replaced `unidecode` (GPLv2) with `anyascii` (MIT)
* Removed `pandoc`, `poppler`, and `tbb` from conda environment dependencies

## Related Issue / Discussion

There are still some final issues with the licensing for `cartopy` which changed from LGPL to BSD-3-Clause in version 0.23. This update will force us to drop Python3.8. I'll address this in a new Pull Request.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -68,6 +68,7 @@ autoapi_options = ["members", "undoc-members", "private-members"]
 # inherit without having to test which work which do not.
 autodoc_mock_imports = [
     "affine",
+    "anyascii",
     "bottleneck",
     "cairo",
     "cartopy",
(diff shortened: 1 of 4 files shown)


huard (approved):

tlvu:
I am guessing a matching dependencies cleanup PR will also be needed on conda packaging side?

Zeitsperre:
> Are `pandoc`, `poppler`, and `tbb` being dropped also due to licensing issues or they are unused? The reason is not apparently clear to me.

As far as I can tell, these dependencies are not needed to run the library (in addition to being under GPL/copyleft licenses):
 - `pandoc` is needed to convert documentation to other formats (this can be done already in ReadTheDocs/sphinx directly)
 - `tbb` and `poppler` are never called in the code

The changes will eventually be ported to the conda-forge package as well!

PR #509, opened 2024-05-21
Update security checks

Zeitsperre:
## Overview

* Added custom endpoints for security in `main.yml`
* Now using commit hashes for specifying action versions

## Additional Information

I think `raven` will be ready for release after this PR

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,7 +3,7 @@ name: raven-wps Testing
 on:
   push:
     branches:
-      - master
+      - main
   pull_request:
 
 env:
@@ -14,6 +14,9 @@ concurrency:
   group: "${{ github.workflow }}-${{ github.ref }}"
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
+permissions:
+    contents: read
+
 jobs:
   black:
     name: Code linting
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -3,7 +3,7 @@ name: raven-wps Testing
 on:
   push:
     branches:
-      - master
+      - main
```
tlvu, line 6: When will this default branch name change be "live"?  You will take care to sync with the changes here https://github.com/bird-house/finch/issues/355#issuecomment-2096900447 to ensure the notebook autodeploy and nightly Jenkins is not broken?

tlvu (approved):
Looks good to me but please also ensure tuto notebook autodeploy to PAVICS and nightly Jenkins are not broken with this change.  Take a look at my comment on Finch repo for the same default branch name change.

.github/workflows/main.yml
Zeitsperre: This has been live for Raven for some time now. I imagine that there aren't any problems.

tlvu: Ah true the branch switch has been already done for Raven, I forgot.  So this github CI workflow change was missing to complete the switch.

All good then.

PR #569, opened 2025-06-19
Use pooch testdata fetching and modernize

Zeitsperre:
## Overview

Changes:

* Replaced testdata fetching mechanism with `pooch`-based caching system
* Updated tooling to replace a few checks with `ruff`, `Makefile` linting checks are now more precise
* Synchronized dependencies and added a few base versions for consistency
* Removed obsolete `setup.cfg`
* `raven` now uses Trusted Publisher for TestPyPI and PyPI releases

## Related Issue / Discussion

This PR is part of an effort to move the `raven-testdata` repo to a versioned scheme that will make it much easier to track changes between major releases of `RavenWPS` and `RavenPy`. It also provides much more thread-safety and prevents needless requests to GitHub when running `pytest`.

I'm pinning things so that this should be the last release to support Python 3.9. We should drop that version in the next significant release (if we do another one).

## Additional Information

https://github.com/CSHS-CWRA/RavenPy/pull/513
https://github.com/Ouranosinc/raven-testdata/pull/43


tlvu (approved):
I guess I can only approve the spirit of this PR.  Don't know much about the test framework on Raven.

Minor comment about whether this affects the notebooks on PAVICS.

PR #570, opened 2025-06-20
adjust docker testing setup

Zeitsperre:
## Overview

Changes:

* Added the `step-security/harden-runner` action to the `docker-testing.yml` workflow
* Synchronized the setup with that of `bird-house/finch`

## Related Issue / Discussion

It would be interesting if we can leverage this workflow to perform docker-build and dockerhub deployment actions

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,3 +25,7 @@ RUN pip install . --no-deps
 EXPOSE 9099
 
 CMD ["gunicorn", "--bind=0.0.0.0:9099", "-t 60", "raven.wsgi:application"]
+# docker build -t pavics/raven .
+# docker run -p 9099:9099 pavics/raven
+# http://localhost:9099/wps?request=GetCapabilities&service=WPS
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -9,24 +10,18 @@ WORKDIR /code
...
 # Start WPS service on port 9099 on 0.0.0.0
 EXPOSE 9099
 
-CMD ["gunicorn", "--bind=0.0.0.0:9099", "raven.wsgi:application"]
-#CMD ["exec raven-wps start -b '0.0.0.0' -c etc/demo.cfg"]
-
-# docker build -t pavics/raven .
-# docker run -p 9099:9099 pavics/raven
-# http://localhost:9099/wps?request=GetCapabilities&service=WPS
-# http://localhost:9099/wps?request=DescribeProcess&service=WPS&identifier=all&version=1.0.0
```
tlvu, line 32: Why remove those manual step comment?  I think it's a reminder for those who wants to test locally on his machine, without having to push bogus commits just to trigger Github pipeline.

tlvu (approved):
LGTM, minor comment

PR #580, opened 2025-08-15
Fixup lint check configuration

Zeitsperre:
## Overview

Changes:

* Added some logic to skip lint checks on `main` (some hooks fail when run on `main` by design)
* Always run `conda` step on pushes to `main` but run it dependent on `lint` in pull requests.
* Update `pre-commit` hooks


tlvu (approved):
Rubber stamp from me but it looks good.

PR #605, opened 2026-04-10
Cookiecutter update

Zeitsperre:
## Overview

Changes:

* Updated project from (relatively outdated) cookiecutter template (cookiecutter-birdhouse):
  * `bump-my-version` has been configured for patch bumps
  * Labels are now applied
  * Python CI libraries now use fixed hashes
  * Several small `ruff` code style violations have been addressed
  * `LOGGER` is now `logger`. No more root-level logging
  * `pylint` and `vulture` have now been configured to run on codebase

## Related Issue / Discussion

The `cookiecutter-birdhouse` template is somewhat old, but many changes adopted there were ported with relative ease here. It might be worthwhile to spend some time on the template so that other birds can benefit.

## Additional Information

While the `tox` configuration is severely out-of-date, we don't test using it, so no issues. Maybe worth considering?


Zeitsperre:
@huard This is ready to go as is, but there are a bunch of changes in https://github.com/bird-house/cookiecutter-birdhouse/pull/162 that this could also benefit from as well.

Either way, RavenWPS docker image could use a refresh.

tlogan2000 (approved):

PR #617, opened 2026-07-22
Docker Autobuild

Zeitsperre:
## Overview

Changes:

* Added a Docker build and Docker Hub publishing workflow
* Dockerfile image adjustments:
    * Changed the base Dockerfile image from `condaforge/mambaforge` to `condaforge/miniforge3`
    * Adjusted the Dockerfile metadata to use `org.opencontainers` metadata conventions
* Fixed the Zenodo DOI badge to use the `latest` URL

## Related Issue / Discussion

Docker Hub Autobuild is a deprecated feature and we need to move away from it before April 2027 (ideally, much sooner): https://docs.docker.com/docker-hub/repos/manage/builds/migrate/

## Additional Information

https://github.com/Ouranosinc/jupyterhub/pull/42
https://github.com/bird-house/finch/pull/549


tlvu (approved):
LGTM.  Same workflow as the https://github.com/Ouranosinc/jupyterhub/pull/42?

Meaning
* `bump-my-version {major | minor | patch | date | build}` or `bump-my-version bump --new-version {#.#.#-YYYYmmdd | #.#.#-YYYYmmdd.b}` support?
* if a tag is seen, the Docker image built on Github will also be pushed to DockerHub?

Zeitsperre:
>     * `bump-my-version {major | minor | patch | date | build}` or `bump-my-version bump --new-version {#.#.#-YYYYmmdd | #.#.#-YYYYmmdd.b}` support?

We don't use `date` here, so that's not supported. `--new-version` is supported but should almost never be required unless a maintainer has really botched something. Standard `bump` calls should be enough.

>     * if a tag is seen, the Docker image built on Github will also be pushed to DockerHub?

Yes, any newly-pushed tags will trigger a deployment here.

tlvu:
> Standard `bump` calls should be enough.

Ok so only `bump-my-version {major | minor | patch | build}`.   Is that in some README file?  Just make sure the instructions there are up-to-date.

> Yes, any newly-pushed tags will trigger a deployment here.

Tag created by `bump-my-version`?  Great !

Zeitsperre:
> > Standard `bump` calls should be enough.
> 
> Ok so only `bump-my-version {major | minor | patch | build}`. Is that in some README file? Just make sure the instructions there are up-to-date.

I just realized that we also don't support `build` at the moment here. Should we? Would be very easy to implement.

This information is in `dev_guide.rst`.

> > Yes, any newly-pushed tags will trigger a deployment here.
> 
> Tag created by `bump-my-version`? Great !

It is not. We rely on maintainers to push tags when a release is ready here. This project isn't as automated as `birdhouse-deploy`. Do we want it to be?

tlvu:
> I just realized that we also don't support `build` at the moment here. Should we? Would be very easy to implement.

If dead simple, why not.  Let's standardize our tagging workflow.

>> Tag created by bump-my-version? 

> It is not. We rely on maintainers to push tags when a release is ready here. 

Is this the current workflow?  Is yes, then maybe there is a reason for that ?

> This project isn't as automated as `birdhouse-deploy`. Do we want it to be?

Same answer, if dead simple, why not.  Let's standardize our tagging workflow.

Zeitsperre:
> Is this the current workflow? Is yes, then maybe there is a reason for that ?

The reason is that designed it that way a million years ago :laughing: I think we can safely update it.

PR #618, opened 2026-08-12
Cruft update

Zeitsperre:
## Overview

Changes:

* Updated the cookiecutter to the latest changes
* Replace `pre-commit` with `prek`
* Drop support for Python 3.10
* More consistent docuemntation formatting
* Project configuration now uses PEPs 639 and 735
* Employ `pytest` v9.0+ conventions
* `tox.ini` replaced by `tox.toml`

## Related Issue / Discussion

This does not yet include the changes found in https://github.com/bird-house/cookiecutter-birdhouse/pull/168

Some changes here may get ported there, particularly the Docker-related ones. I can add these at a later time.

## Additional Information

The cookiecutter is nowhere near as modern as the Ouranosinc project. There are likely a multitude of things that would be worthwhile porting from there as well.


docs/source/apidoc/modules.rst
aulemahal, line 1: Did you mean to push the `apidoc` files ? I just know we don't do it in xclim.

https://github.com/Ouranosinc/xclim/blob/879dcf2482bd6ba3bca6653bcb45f21c707524bb/.gitignore#L109-L114

aulemahal (approved):

docs/source/apidoc/modules.rst
Zeitsperre: Good catch. In `xclim`, we execute apidoc during the build process for ReaTheDocs: https://github.com/Ouranosinc/xclim/blob/879dcf2482bd6ba3bca6653bcb45f21c707524bb/.readthedocs.yml#L17

We also do this in the Ouranos cookiecutter, but not the Birdhouse one. Perhaps we should?

